PR #7, opened 2021-06-28
Add 'Show / Create / Revoke token' feature

Silvan-WMDE:
This adds protected web routes for showing, creating and revoking API tokens. The token needs to be included as `Bearer` token in the `Authorization` header field and can be tested on the `/api/user/` endpoint.

Bug: [T285330](https://phabricator.wikimedia.org/T285330)


guergana (dismissed):
Good work! We probably need to squash all the commits into one.

resources/views/showToken.blade.php
```diff
@@ -0,0 +1,30 @@
...
+            <div id="tokenId">ID: {{ $token->id }}</div>
+            <div id="tokenName">name: {{ $token->name }}</div>
+            <div id="tokenAbilities">abilities: {{ implode(', ', $token->abilities) }}</div>
+            <div id="tokenCreated">created at: {{ $token->created_at }}</div>
+            <div id="tokenLastUsed">last used at: {{ $token->last_used_at }}</div>
+            <div id="tokenRevoke"><a href="{{ route('token.revoke', [ 'id' => $token->id ] ) }}">Revoke</a></div>
+        </div>
+        <hr/>
+    @endforeach
+
```
itamargiv, line 27: Just for the sake of completion, but for sure not a  blocker:

```suggestion
<div id="back"><a href="/">home</a></div>
```

itamargiv (dismissed):
Nicely done, thank you! :bow:  We will need to actually perform all the migrations we added in the live server :)

itamargiv (approved):
:100:

PR #9, opened 2021-06-30
Change database charset from utf8mb4 to utf8

Silvan-WMDE:
According to Wikitech Help [1], this avoids a database error on toolforge,
which says 'Index column size too large.'

[1] https://wikitech.wikimedia.org/wiki/Help:Toolforge/Database#User_databases


itamargiv (approved):
Thank you!

PR #10, opened 2021-07-01
Add php linting with phpcs

guergana:
First commit: 
- install phpcs locally
- add configuration file
- update developer documentation

Second commit 
- fixed linting errors

Third commit
- add github action for linting with phpcs

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,13 @@
+name: Lint
+
+on: [push]
+
+jobs:
+  phpcs-lint:
+    runs-on: ubuntu-latest
+    steps:
+        - uses: actions/checkout@v2
+        - name: PHP Lint
+          uses: php-actions/composer@v6
+          with:
+              command: run lint
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
         - uses: actions/checkout@v2
-        - name: PHP Lint
+        - name: Install dependencies
           uses: php-actions/composer@v6
-          with:
-              command: run lint
+        - name: PHP Lint
+          run: "composer run lint"
--- a/docs/README.md
+++ b/docs/README.md
@@ -10,6 +10,7 @@
   - [Stop the application server](#stop-the-application-server)
   - [Destroy all the things](#destroy-all-the-things)
 - [Working with OAuth](#oauth)
+- [PHP Linting](#phpcs)
 - [Troubleshooting](#troubleshooting)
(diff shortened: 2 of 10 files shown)


docs/README.md
```diff
@@ -129,6 +130,16 @@ In production, this application relies on wikidata.org's OAuth capabilities in o
 
 1. As soon as you receive the email from the WMF team that your consumer is approved, you may start testing your application by logging in through your local instance's home page.
 
+## PHP Linting <a id="phpcs"></a>
+
+The application uses `phpcs` to detect code format violations.
+
+To run phpcs: `./vendor/bin/phpcs`
```
itamargiv, line 137: This is great! Thanks for updating the docs.

I think it might be easier if we also added a `lint` script to `composer.json` and then use that one in our GH action and the docs instead of calling the phpcs binary directly. That way we will be free to switch tools if we see the need to do so, without having to update everything again.

Also, this would mean we would be able to provide a more consistent development experience by documenting that commands `sail composer run lint` and `sail composer run fix` here instead.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,14 @@
...
+
+jobs:
+  phpcs-lint:
+    runs-on: ubuntu-latest
+    steps:
+        - uses: actions/checkout@v2
+        - name: PHP Lint
+          run: |
+              composer install
+              ./vendor/bin/phpcs
```
itamargiv, line 13: I think we can use `php-actions/composer@v6` for the installation step, and if we add a composer script for linting we can run it like so:

```suggestion
        - uses: actions/checkout@v2
        - name: PHP Lint
          uses: php-actions/composer@v6
          with:
              command: lint
```

itamargiv (changes requested):
Looks great! Thanks for the implementation and the fixes, I have a few minor comments, and then we're good to go

docs/README.md
guergana: Very true. Will update.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,13 @@
...
+
+jobs:
+  phpcs-lint:
+    runs-on: ubuntu-latest
+    steps:
+        - uses: actions/checkout@v2
+        - name: PHP Lint
+          uses: php-actions/composer@v6
+          with:
+              command: run lint
```
itamargiv, line 13:
```suggestion
        - name: Install dependencies
          uses: php-actions/composer@v6
        - name: PHP Lint
          uses: php-actions/composer@v6
          with:
              command: run lint
```

itamargiv (approved):
\o/ Woohoo, thank you @guergana!

PR #11, opened 2021-07-05
Add file upload feature

Silvan-WMDE:
This will accept a file upload via multipart/form POST request at the `/api/upload` endpoint. The request must contain a `mismatchFile` field with an attached csv file and may optionally contain a name and a description field.

Upload validation will check that:
1. the user has sufficient privileges to upload a mismatch file,
2. a mismatch file is provided,
3. the file is not larger than 10MB,
4. the file is in text/csv format.


app/Http/Controllers/UploadController.php
```diff
@@ -0,0 +1,37 @@
...
+            return response(['success' => false, 'reason' => 'User has no upload privilege' ], 403);
+        }
+
+        $request->validate([
+            'mismatchFile' => [ 'required', 'file', 'max:10000', 'mimes:csv,txt' ]
+        ]);
+
+        $uploadName = $request->name;
+        $description = $request->description;
+        $filename = date('Ymd_His') . '-mismatch-upload.' . $request->user()->mw_userid . '.csv';
```
guergana, line 27: date works better. good one.

app/Http/Controllers/UploadController.php
```diff
@@ -0,0 +1,37 @@
...
+namespace App\Http\Controllers;
+
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\File;
+
+class UploadController extends Controller
+{
+    public function __construct()
+    {
+        $this->middleware('auth')->only('upload');
```
mariushoch, line 12: I'm a little confused… what does this do? Does the auth middleware still know anything about uploads at this point?

app/Http/Controllers/UploadController.php
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        $request->validate([
+            'mismatchFile' => [ 'required', 'file', 'max:' . env('UPLOAD_SIZE_LIMIT'), 'mimes:csv,txt' ]
+        ]);
+
+        $uploadName = $request->name;
+        $description = $request->description;
+        $filename = date('Ymd_His') . '-mismatch-upload.' . $request->user()->mw_userid . '.csv';
+        $request->file('mismatchFile')->storeAs('mismatch-files', $filename);
```
mariushoch, line 28: (Probably for a follow-up:) We should mount this into the container as a volume, via docker-compose.

app/Http/Controllers/UploadController.php
Silvan-WMDE: You're right - this is not needed. Must be a copy-paste error. Since this is an API endpoint, auth happens via the sanctum token, which is defined in `routes/api.php` like this: `Route::middleware('auth:sanctum')->post('/upload' ...`.

Removed.

app/Http/Controllers/UploadController.php
Silvan-WMDE: To be honest, we don't even know what we are going to do with the uploaded file post import. Maybe we will just delete it?

mariushoch: Could be useful for debugging purposes, but maybe I'm over engineering this already.

mariushoch (approved):
Manually tested.

PR #12, opened 2021-07-06
Manage allowed upload users via artisan console

Silvan-WMDE:
This change adds two custom artisan commands: `uploadUsers:show` and `uploadUsers:set {file}`. The first one shows the list of users who are allowed to upload mismatch files to the store. The second one takes the name of an allowlist file as an argument and replaces the current list of users with the one from the specified file. The allow list file must be placed in the `storage/app/allowlist` directory. Users who are sending API requests to the `/api/upload/` endpoint will consequently be checked against this list.

Documentation of this feature is added to a new Administration Guide.

Bug: [T286024](https://phabricator.wikimedia.org/T286024)

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,10 @@ A tool to review mismatches between Wikidata and External Databases
 ## User Guide
 
  * See the [User Guide](docs/UserGuide.md) for instructions on how to use Wikidata Mismatch Finder.
+ 
+## Administration Guide
+
+ * See the [Administration Guide](docs/AdminGuide.md) for instructions on how to run administration tasks.
 
 ## Local Development
 
@@ -13,9 +17,3 @@ A tool to review mismatches between Wikidata and External Databases
 * Stop the application server with `sail down`
 * Make sure to setup a local testing oauth consumer by following [these instructions](docs/README.md#oauth)
 * If you encounter any issues, see the [troubleshooting section](docs/README.md#troubleshooting).
-
-## User Guide
- * See the [User Guide](docs/UserGuide.md) for instructions on how to use Wikidata Mismatch Finder.
-
-## Administration Guide
- * See the [Administration Guide](docs/AdminGuide.md) for instructions on how to run administration tasks.
--- a/README.md
+++ b/README.md
@@ -13,3 +13,9 @@ A tool to review mismatches between Wikidata and External Databases
 * Stop the application server with `sail down`
 * Make sure to setup a local testing oauth consumer by following [these instructions](docs/README.md#oauth)
 * If you encounter any issues, see the [troubleshooting section](docs/README.md#troubleshooting).
+
+## User Guide
+ * See the [User Guide](docs/UserGuide.md) for instructions on how to use Wikidata Mismatch Finder.
+
+## Administration Guide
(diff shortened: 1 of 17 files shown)


app/Console/Commands/SetUploadUsers.php
```diff
@@ -0,0 +1,64 @@
...
+            echo "File not found\n";
+            return 1;
+        }
+
+        UploadUser::truncate();
+
+        // read file from local disc, split by newline and filter out empty lines
+        $lines = array_filter(
+            explode("\n", Storage::disk('local')->get($allowlistFile))
+        );
```
mariushoch, line 55: This is very convenient, but array_filter will also filter out "0". To make this more reliable, I would suggest to not use it, but rather `trim( $line ) === ''` (for each line).

```
php > print_r(array_filter(["0"]));
Array
(
)
```

app/Console/Commands/SetUploadUsers.php
```diff
@@ -0,0 +1,64 @@
...
+     * @var string
+     */
+    protected $signature = 'uploadUsers:set {allowList}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Set the list of users who are allowed to upload mismatch files by providing a txt file;';
```
mariushoch, line 23: This should probably say where the text file should be located/ how it should be provided.

app/Console/Commands/SetUploadUsers.php
Silvan-WMDE: tbh, I still prefer the convenience of array_filter over trimming each line one-by-one. Stackoverflow has an [elegant solution to the "0" problem](https://stackoverflow.com/a/38081606), but I would really find it rather strange to include the "strlen" callback. If the next reviewer asks what it is for, I'd have to add a comment to the code to explain that this will allow the username "0", which I am pretty sure will never show up on our allow list.

app/Console/Commands/SetUploadUsers.php
Silvan-WMDE: I added a description to the `allowlist` input argument. Try `sail artisan help uploadUsers:set` to see the help message.

guergana (dismissed):
Great work. Also great documentation.

README.md
```diff
@@ -9,3 +9,9 @@ A tool to review mismatches between Wikidata and External Databases
 * Stop the application server with `sail down`
 * Make sure to setup a local testing oauth consumer by following [these instructions](docs/README.md#oauth)
 * If you encounter any issues, see the [troubleshooting section](docs/README.md#troubleshooting).
+
+## User Guide
```
guergana, line 13: :sunglasses:  cool beans

itamargiv (dismissed):
Looks good so far! Thanks for implementing this. There are still a couple of comments to address, and I think it might also be worthwhile to try and create a seeding script of at least one user which is in both the `User` and the `UploadUser` tables, to make local development a little nicer.

itamargiv (commented):
The ball is in your court @guergana and @mariushoch

guergana (approved):
Approved by @itamargiv and I.

PR #14, opened 2021-07-12
Integration Testing (Container Duo)

itamargiv:


guergana:
Closing because @itamargiv and I have decided to go with the SQLite approach. 

Reasoning here:

Alright, here are my 2 cents on the integration tests issue:
1) I think for convenience and ease of use in local development, we should go with SQLite variation at first. This will also make on PR check a cinch, since we don't need to mess around with extra containers. Also, this should cover 95% of the cases.
2) For the extra 5% (god I hate numbers that are pulled out of nowhere) and our own sense of security, I think we can set up an extra CI Job to run only on GH where we use the miraculous GH secrets to spin up a version of the container that is already included with sail, without the need to define an extra one for testing. If we want this to be reproducible locally, we can add instructions on how to spin up an additional maria db instance for testing using docker (or even create our own command for it, and for running the second variation of the integration tests)

PR #15, opened 2021-07-12
Integration Testing (SQLite Mock)

itamargiv:


tests/Feature/ApiRouteTest.php
```diff
@@ -0,0 +1,108 @@
...
+
+        $user = User::factory()->create();
+
+        Storage::fake('mismatchFiles');
+        $file = UploadedFile::fake()->create('mismatchFile.csv');
+
+        Sanctum::actingAs($user);
+        
+        $response = $this->post('/api/upload', ['mismatchFile' => $file]);
+
```
itamargiv, line 83: Looks great! We probably also want to check that the file was actually created in the fake storage after the upload, as part of this test. Take a look at the [laravel docs](https://laravel.com/docs/8.x/http-tests#testing-file-uploads) for a general example, and also at [this commit](https://github.com/wmde/wikidata-mismatch-finder/pull/12/commits/1bfdd36b50414cfabf71faf6a636d1b2796c5d2c) to see how to configure a storage "disk" you can later fake in the tests (inspired by a [stack overflow answer](https://stackoverflow.com/questions/54039755/adding-custom-path-for-storage-in-filesystems-php-laravel#54041057)).

tests/Feature/WebRouteTest.php
```diff
@@ -0,0 +1,24 @@
...
+     *
+     *  @return void
+     */
+    // phpcs:ignore
+    public function test_welcome_route()
+    {
+        $response = $this->get('/');
+
+        $response->assertStatus(200);
+        $response->assertViewIs('welcome');
```
itamargiv, line 22: This is so magnificently simple, I love it :)

itamargiv (commented):
Nicely done! I'm still missing a test for the three remaining auth routes, but I think we can put them in another PR. There's one last thing to be done here, and then we're good to go. (Also, I cannot approve this, since I created the PR :grimacing: )

itamargiv:
Oh, and one last thing, we need to update the GitHub action to run the tests, give me a shout if you need help doing that, I'm always happy to contrib.

tests/Feature/WebRouteTest.php
guergana: It's magic :sparkler: :sparkler:

tests/Feature/ApiRouteTest.php
guergana: You are right, I thought at some point about adding it but forgot down the road.

itamargiv (commented):
TOTALLY approved! (only… I can't really approve because I opened the PR)

guergana (approved):
Awesum!

PR #17, opened 2021-07-15
Copy uploaderlist to storage

guergana:
Add a step in the deploy workflow to avoid losing the uploaderslist everytime. 

Bug: T286024


itamargiv (approved):
\o/

PR #20, opened 2021-07-16
Use rsync to deploy, don't delete and re-copy

mariushoch:
In order for this to work reliably, we make sure that the deployed files are group-writable, then change the new source files to have `tools.mismatch-finder` as group. After that, we `rsync` the new files in place.
Finally, we run `take` to make sure all (new) files/folders are owned by tools.mismatch-finder. Due to `take`'s symlink handling, we need to run it on for all files/folders, and can't let it recurse on its own.

https://phabricator.wikimedia.org/T286056

### DIFF
--- a/.github/workflows/deploy-app.yml
+++ b/.github/workflows/deploy-app.yml
@@ -1,7 +1,7 @@
 name: Build and Deploy to Toolforge
 
-on: 
-  push: 
+on:
+  push:
    branches:
     - main
     - gh-actions/**
@@ -23,8 +23,8 @@ jobs:
         version: 2
         php_version: 7.3
     - name: Installing node dependencies
-      run: npm ci    
-    - name: Minify and Build CSS and JS 
+      run: npm ci
+    - name: Minify and Build CSS and JS
       run: "npm run prod"
     - name: Syncing code to Toolforge
       uses: appleboy/scp-action@master
@@ -45,9 +45,10 @@ jobs:
         key: ${{ secrets.KEY }}
         passphrase: ${{ secrets.PASSPHRASE }}
         script: |
-          become mismatch-finder rm -rf mismatch-finder-repo
-          cp -rv mismatch-finder-repo/ ~tools.mismatch-finder/
-          become mismatch-finder take mismatch-finder-repo
-          become mismatch-finder cp .env mismatch-finder-repo/
-          become mismatch-finder mkdir -p mismatch-finder-repo/storage/app/allowlist
-          become mismatch-finder cp uploaders.txt mismatch-finder-repo/storage/app/allowlist/
+          # Make sure ~tools.mismatch-finder/mismatch-finder-repo is group writable
+          become mismatch-finder chmod -R g+rwx ~tools.mismatch-finder/mismatch-finder-repo
+          # Change group of ~/mismatch-finder-repo (including symlinks) to tools.mismatch-finder
+          chgrp --no-dereference -R tools.mismatch-finder ~/mismatch-finder-repo
+          rsync -rlgD --delete --delay-updates --exclude '.nfs*' --exclude mismatch-finder-repo/.env --exclude mismatch-finder-repo/storage/app/ ~/mismatch-finder-repo ~tools.mismatch-finder/
+          # take aborts recursion whenever it encounters a symlink, thus we use find+xargs to make sure all folders and file ares handled.
+          find ~tools.mismatch-finder/mismatch-finder-repo -type d,f \! -user tools.mismatch-finder -print0 | become mismatch-finder xargs -r --null take 2>&1 | { grep -vF 'will not follow or touch symlinks' || true; }


itamargiv:
The command looks great, but I think we have an issue with some permissions. tested using a `gh-actions/*` branch https://github.com/wmde/wikidata-mismatch-finder/runs/3087738511

mariushoch:
I just realized this wont work… I'll have another look at this later.

mariushoch:
Test run: https://github.com/wmde/wikidata-mismatch-finder/runs/3109603132?check_suite_focus=true

mariushoch:
@itamargiv These are owned by you, can you have a look and chown them to the tool?
```
$ find /data/project/mismatch-finder/mismatch-finder-repo -user itamar -perm 0644
/data/project/mismatch-finder/mismatch-finder-repo/.git/refs/heads/gh-actions/deploy-w-rsync
/data/project/mismatch-finder/mismatch-finder-repo/.git/refs/remotes/origin/gh-actions/deploy-w-rsync
/data/project/mismatch-finder/mismatch-finder-repo/.git/logs/refs/heads/gh-actions/deploy-w-rsync
/data/project/mismatch-finder/mismatch-finder-repo/.git/logs/refs/remotes/origin/gh-actions/deploy-w-rsync```

**Comment by itamargiv** [comment]
> @itamargiv These are owned by you, can you have a look and chown them to the tool?
> 
> ```
> $ find /data/project/mismatch-finder/mismatch-finder-repo -user itamar -perm 0644
> /data/project/mismatch-finder/mismatch-finder-repo/.git/refs/heads/gh-actions/deploy-w-rsync
> /data/project/mismatch-finder/mismatch-finder-repo/.git/refs/remotes/origin/gh-actions/deploy-w-rsync
> /data/project/mismatch-finder/mismatch-finder-repo/.git/logs/refs/heads/gh-actions/deploy-w-rsync
> /data/project/mismatch-finder/mismatch-finder-repo/.git/logs/refs/remotes/origin/gh-actions/deploy-w-rsync
> ```

@mariushoch this is what the `become mismatch-finder take` command is for, transferring ownership of the files to the tool. see https://wikitech.wikimedia.org/wiki/Portal:Toolforge/Tool_Accounts#Take_ownership_of_files

**Comment by mariushoch** [comment]
> @mariushoch this is what the `become mismatch-finder take` command is for, transferring ownership of the files to the tool. see https://wikitech.wikimedia.org/wiki/Portal:Toolforge/Tool_Accounts#Take_ownership_of_files

Sadly that fails in this specific case:
```
$ become mismatch-finder take mismatch-finder-repo/.git/refs/heads/gh-actions/deploy-w-rsync
mismatch-finder-repo/.git/refs/heads/gh-actions/deploy-w-rsync: You need to share a group with the file
```

**Comment by itamargiv** [comment]
> > @mariushoch this is what the `become mismatch-finder take` command is for, transferring ownership of the files to the tool. see https://wikitech.wikimedia.org/wiki/Portal:Toolforge/Tool_Accounts#Take_ownership_of_files
> 
> Sadly that fails in this specific case:
> 
> ```
> $ become mismatch-finder take mismatch-finder-repo/.git/refs/heads/gh-actions/deploy-w-rsync
> mismatch-finder-repo/.git/refs/heads/gh-actions/deploy-w-rsync: You need to share a group with the file
> ```

Oh, I think I have an idea why this fails, let me try something, then I'll make a suggestion if it works

**Comment by itamargiv** [comment]
Alright, after some asking around, there doesn't seem to be an easy solution for these permission issues. The way I see it, we have two options:

1) If these git files are the only issue in our way, I would suggest ignoring the `.git` folder altogether, since we don't need it in the tool account itself (could save us some space too).

2) The second option is to just `cp` the repository over to some `aux` folder, take that folder as the tool, perform the rsync as the tool - and delete the aux - as the tool :smile:. Though this is a more resource heavy, time-consuming operation :grimacing: .

File: .github/workflows/deploy-app.yml [hunk 1]
```diff
@@ -39,9 +39,7 @@ jobs:
         key: ${{ secrets.KEY }}
         passphrase: ${{ secrets.PASSPHRASE }}
         script: |
-          become mismatch-finder rm -rf mismatch-finder-repo
-          cp -rv mismatch-finder-repo/ ~tools.mismatch-finder/
+          # Make sure ~tools.mismatch-finder/mismatch-finder-repo is group writable
+          become mismatch-finder chmod -R g+rwx ~tools.mismatch-finder/mismatch-finder-repo
+          rsync -rltD --delete --delay-updates --exclude mismatch-finder-repo/.env --exclude mismatch-finder-repo/storage/app/allowlist/uploaders.txt ~/mismatch-finder-repo ~tools.mismatch-finder/
```
itamargiv, line 44: One more thing, we would probably want to keep all of `storage/app` intact for now, as we also upload files to a subdirectory there.

guergana (commented):
> Alright, after some asking around, there doesn't seem to be an easy solution for these permission issues. The way I see it, we have two options:
> 
>     1. If these git files are the only issue in our way, I would suggest ignoring the `.git` folder altogether, since we don't need it in the tool account itself (could save us some space too).
> 
>     2. The second option is to just `cp` the repository over to some `aux` folder, take that folder as the tool, perform the rsync as the tool - and delete the aux - as the tool smile. Though this is a more resource heavy, time-consuming operation grimacing .

We could alternatively just leave it as it was and not use `rsync`

itamargiv:
> We could alternatively just leave it as it was and not use `rsync`

Well, not really, since, if we leave it as-is we will lose the upload files, since they are stored in a subfolder of our repo.

mariushoch:
Test run: https://github.com/wmde/wikidata-mismatch-finder/runs/3132728947?check_suite_focus=true

.github/workflows/deploy-app.yml
mariushoch: Sounds good, changed.

.github/workflows/deploy-app.yml
```diff
@@ -39,9 +39,10 @@ jobs:
...
-          become mismatch-finder mkdir -p mismatch-finder-repo/storage/app/allowlist
-          become mismatch-finder cp uploaders.txt mismatch-finder-repo/storage/app/allowlist/
\ No newline at end of file
+          # Make sure ~tools.mismatch-finder/mismatch-finder-repo is group writable
+          become mismatch-finder chmod -R g+rwx ~tools.mismatch-finder/mismatch-finder-repo
+          # Change group of ~/mismatch-finder-repo (including symlinks) to tools.mismatch-finder
+          chgrp --no-dereference -R tools.mismatch-finder ~/mismatch-finder-repo
+          rsync -rlgD --delete --delay-updates --exclude '.nfs*' --exclude mismatch-finder-repo/.env --exclude mismatch-finder-repo/storage/app/ ~/mismatch-finder-repo ~tools.mismatch-finder/
+          # take aborts recursion whenever it encounters a symlink, thus we use find+xargs to make sure all folders and file ares handled.
+          find ~tools.mismatch-finder/mismatch-finder-repo -type d,f \! -user tools.mismatch-finder -print0 | become mismatch-finder xargs -r --null take 2>&1 | { grep -vF 'will not follow or touch symlinks' || true; }
```
itamargiv, line 48: This line confuses me a bit. I'm not sure why we would need to follow symlinks in this case? Did it cause nay issues before?

itamargiv (commented):
Looking great! I have one question, but other than that I'm happy it works. Thank you!!

.github/workflows/deploy-app.yml
mariushoch: We don't need to follow symlinks… but the problem is that `take` stops recursing through the directories in case it encounters a symlink (it doesn't just skip it). So we can't just run take on all of `mismatch-finder-repo` at once, as it wont process all files.

itamargiv: Gotcha, thanks!

itamargiv (approved):
Great work! Well done on the research, and for finding a workaround to these nagging permission issues. :clap: :+1:

itamargiv:
Hey, sorry I had to dismiss this review, I got the following email from toolforge on Saturday, about them having to terminate the `rsync` process:

Screenshot from 2021-07-26 10-51-14

itamargiv:
Rerunning the test on a `gh-actions/*` branch to verify that the process ends

itamargiv:
Looks like the process ends, but seems like we're back to having permission issues? Perhaps which should go back to the drawing board and reconsider this rsync approach.

itamargiv (approved):
Merci merci merci

PR #21, opened 2021-07-19
Create mismatch import job

itamargiv:
This PR introduces the `Mismatch` model, migration, factory and seeder. As well as [jobs](https://laravel.com/docs/8.x/queues) to validate a CSV file and import mismatches from a file and into our mismatches' database table.

### DIFF
--- a/app/Http/Controllers/ImportController.php
+++ b/app/Http/Controllers/ImportController.php
@@ -9,6 +9,8 @@ use Illuminate\Support\Facades\Gate;
 use App\Http\Resources\ImportMetaResource;
 use Illuminate\Http\Resources\Json\JsonResource;
 use App\Jobs\ImportCSV;
+use Illuminate\Support\Facades\Bus;
+use App\Jobs\ValidateCSV;
 
 class ImportController extends Controller
 {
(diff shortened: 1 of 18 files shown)


app/Http/Controllers/ImportController.php
```diff
@@ -70,6 +73,11 @@ public function store(Request $request): JsonResource
 
         $meta->save();
 
+        Bus::chain([
```
guergana, line 76: What is a Bus?? O: the new things keep on coming!!

itamargiv: A bus is essentially (I might be slightly inaccurate here) a mechanism that is responsible for the order in which jobs are put in a queue. In this case, what we are trying to do is to "bus together" the CSV file validation job and the CSV import job. But currently, since this is a WIP - it is unsuccessful, and I'm trying to understand why :disappointed:

itamargiv: Argh, found it! I was using `->map()` instead of `->each()` :roll_eyes:

itamargiv:
Closing this PR in favor of a feature branch, with multiple smaller PRs.

PR #23, opened 2021-07-23
Add Mismatch model, migration, factory and seeder

itamargiv:
This PR introduces the mismatches table and model to our system. This is the first part of a bigger import feature, and therefore, It will be merged into a feature branch rather than main.

### DIFF
--- a/database/migrations/2021_07_19_123858_create_mismatches_table.php
+++ b/database/migrations/2021_07_19_123858_create_mismatches_table.php
@@ -20,7 +20,7 @@ class CreateMismatchesTable extends Migration
             $table->string('property_id');
             $table->binary('wikidata_value'); // Saving as BLOB to enable storing JSON seralizations
             $table->text('external_value');
-            $table->string('external_url', 1024)->nullable();
+            $table->string('external_url', 2048)->nullable();
             $table->enum('status', [
                 'pending',
                 'wikidata',
--- a/database/migrations/2021_07_19_123858_create_mismatches_table.php
+++ b/database/migrations/2021_07_19_123858_create_mismatches_table.php
@@ -1,0 +1,45 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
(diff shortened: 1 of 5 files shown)


database/migrations/2021_07_19_123858_create_mismatches_table.php
```diff
@@ -0,0 +1,45 @@
...
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('mismatches', function (Blueprint $table) {
+            $table->id();
+            $table->string('statement_guid');
+            $table->string('property_id');
+            $table->binary('wikidata_value'); // Saving as BLOB to enable storing JSON seralizations
+            $table->text('external_value');
```
Silvan-WMDE, line 22: Is the text() intended here, insted of string()? The metadata doc says that the field should be length-limited.

database/migrations/2021_07_19_123858_create_mismatches_table.php
```diff
@@ -0,0 +1,45 @@
...
+            $table->string('external_url', 1024)->nullable();
```
Silvan-WMDE, line 23: Here, in contrast, the metadata doc does not say anything about a length limit. I agree that 1024 would be pretty long indeed, but generally URLs up to 2048 seem to be fine.

Silvan-WMDE (approved):
tested locally and works. Two minor comments below.

database/migrations/2021_07_19_123858_create_mismatches_table.php
itamargiv: it is, but I anticipate these fields might be longer than we expect, so I will limit the length at the business logic level rather than the database level.

database/migrations/2021_07_19_123858_create_mismatches_table.php
itamargiv: That is a good point, I will set it to 2048, and use the url limits from wikidata, which are 1500

itamargiv: @Silvan-WMDE done, thank you!

Silvan-WMDE:
OK, then let's get this merged...

PR #27, opened 2021-07-27
Layout fix

guergana:
Fixes to layout commit: 
- h1 font weight set to 400
- padding applied to anchor tabs instead of li element
- svg served as asset

### DIFF
--- a/resources/sass/app.scss
+++ b/resources/sass/app.scss
@@ -40,6 +40,7 @@ header {
         line-height: 40px;
         margin: 0;
         padding: 0 0 28px 0;
+        font-weight: 400;
     }
 
     .top-nav-right {
@@ -79,7 +80,6 @@ nav.tabs {
         border-bottom: 1px solid #A2A9B1;
 
         li {
-            //padding: 4px 10px;
             border: 2px solid white;
             transition-property: border;
             transition-duration: 100ms;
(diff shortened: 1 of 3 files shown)


resources/sass/app.scss
```diff
@@ -93,7 +93,11 @@ nav.tabs {
             }
         
             a {
+                display: block;
                 color: #54595D;
+                padding: 4px 10px;
+                line-height: 24px;
+
```
Silvan-WMDE, line 100: works great now!

resources/sass/app.scss
```diff
@@ -79,7 +80,6 @@ nav.tabs {
         border-bottom: 1px solid #A2A9B1;
 
         li {
-            padding: 4px 10px;
             border: 2px solid white;
             transition-property: border;
```
Silvan-WMDE, line 84:
```suggestion
            transition-property: border-bottom;
```
Without this change, my Chrome browser flashes the borders around the nav tabs upon reload. You can see it more clearly when you set the transition-duration to 1000ms (for testing).

Silvan-WMDE (commented):
one last change - then it should be good. :-)

resources/sass/app.scss
guergana: Thanks for the suggestion. I will commit it as part of other changes I am adding.

guergana: Seems the transitions firing onload is a Chrome bug. Applying [this solution](https://stackoverflow.com/questions/14389566/stop-css-transition-from-firing-on-page-load/42969608#42969608) works.

Silvan-WMDE (approved):
brilliant - works like a charm!

PR #30, opened 2021-07-28
Improve authentication token page

guergana:
This PR adds the GUI for letting users create, delete and regenerate tokens.


resources/views/welcome.blade.php
```diff
@@ -1,11 +1,7 @@
 <x-layout>
-    <p>Coming soon, to a screen near you...</p>
------------------------
     @auth
-        <p>Hello, {{ Auth::user()->username }}! <a href="{{ route('logout') }}">Logout</a></p>
-
-        Check out our <a href="{{ route('token') }}">shiny new API tokens</a>!
+     DONT know how to add the redirect here!
```
itamargiv, line 3: This is not a redirect. I think the page should just display different content for people who are logged in, and this content should be here.

Silvan-WMDE: See https://github.com/wmde/wikidata-mismatch-finder/commit/5247f4f74645311ecfb758b556c0de2b5e16eb3e for an attempt on how it can be done.

guergana: Thanks @Silvan-WMDE . I have merged the commit into this one.

app/Http/Controllers/ApiTokenController.php
```diff
@@ -4,24 +4,36 @@
...
-            return redirect(route('token'));
+            // token already exists and we want to create a new one
+            // we delete the previous tokens
+            Auth::user()->tokens()->delete();
         }
 
         $token = $request->user()->createToken('apiToken');
-        return view('newToken', [ 'newToken' => $token->plainTextToken ]);
+
+        return redirect('auth/token')->with('flashToken', $token->plainTextToken);
```
itamargiv, line 36: Nice! Thanks for listening to my crazy talk :D

app/Http/Controllers/ApiTokenController.php
```diff
@@ -4,24 +4,36 @@
...
     }
 
     public function createToken(Request $request)
     {
         if (sizeof($request->user()->tokens) > 0) {
-            // token already exists
-            return redirect(route('token'));
+            // token already exists and we want to create a new one
+            // we delete the previous tokens
+            Auth::user()->tokens()->delete();
```
itamargiv, line 31: Cool that you also got to implement regeneration!

app/Http/Controllers/ApiTokenController.php
```diff
@@ -4,24 +4,36 @@
...
-use App\Models\User;
+use Illuminate\Support\Facades\Gate;
 
 class ApiTokenController extends Controller
 {
     public function showToken(Request $request)
     {
-        return view('showToken', [ 'tokens' => $request->user()->tokens ]);
+        if (!Auth::check()) {
+            return view('welcome');
```
itamargiv, line 14: Technically, I would also get rid of the welcome view, but this is probably an optimization we can do later.

resources/sass/app.scss
```diff
@@ -1,11 +1,32 @@
 @import url('https://fonts.googleapis.com/css2?family=Lato&family=Source+Serif+Pro&display=swap');
 
 $mobile-max-size: 720px;
+$color-base-0: #000;
```
itamargiv, line 4: Ideally, I would import these from WiKit already, but since this is a side priority, I would definitely agree that we can take care of that later, in another task.

resources/sass/app.scss
```diff
@@ -151,3 +172,129 @@ dl.import-meta dd {
     margin: 0;
     padding-bottom: 16px;
 }
+p {
+    margin: 0 0 16px 0;
+}
+
+.button {
```
itamargiv, line 179: Amazing work! thank you for doing this, it's a lot of CSS to sift through and take from WiKit.

tests/Feature/AuthTokenTest.php
```diff
@@ -58,9 +70,9 @@ public function test_createTokenWhenNoneExists_returnsNewToken()
         $response = $this->actingAs($user)
                          ->get('/auth/createToken');
 
-        $response->assertStatus(200);
-        $response->assertViewIs('newToken');
-        $response->assertViewHas('newToken');
+        $response->assertStatus(302);
+        $response->assertRedirect('auth/token');
+        $response->assertSessionHas('flashToken');
```
itamargiv, line 75: :+1:

app/Http/Controllers/OAuthUserController.php
```diff
@@ -30,6 +30,9 @@ public function callback()
         ]);
 
         Auth::login($user, false);
+        if (Auth::check()) {
```
itamargiv, line 33: > Technically, I would also get rid of the welcome view, but this is probably an optimization we can do later.

Same comment here, we could avoid these checks if we just decided that the token page is our home page. But that really doesn't matter at this point I think, we can optimize this later if needed.

itamargiv (dismissed):
Works like a charm, looks fab! Including regenerate \o/ I left a few comments to consider, but since we can handle these issues later, consider this... Approved! Thank you :100: !

app/Http/Controllers/ApiTokenController.php
guergana: It was easier than I expected. The thing is that in the laravel docs is not clear where you have to put things. But once i figured it out it was all smooth! Thanks!

Silvan-WMDE (approved):
great work - thank you!

PR #31, opened 2021-07-29
Manage homepage view based on login status

Silvan-WMDE:
This change establishes '/auth/token' as the de-facto homepage by
redirecting the '/' web route immediately. This will change as soon
as '/' points to the actual "Mismatch Finder" web page.

At '/auth/token', logged in users will be shown their API token,
whereas unauthenticated users will see the 'welcome' view, asking them
to please log in.


guergana (approved):
Looks great! Thanks!

PR #32, opened 2021-07-29
Fix package lock

itamargiv:
This PR fixes the package lock to be set to only one version of node - the one sail currently uses, version 16. This was done because, while trying to review frontend code, I had to install dependencies, and it completely changed the package lock file (which it shouldn't have done, since I made no changes).

Therefore, the correct package file is now checked in, and the instructions were updated to reflect this issue.

To validate, check this branch out, and run `sail npm install`. There should be no changes to the package JSON file.


Silvan-WMDE (approved):
Checked this branch out, ran sail npm install and there are no changes to the package JSON file. Works.

itamargiv:
Thank You!

PR #35, opened 2021-07-30
[Imports 0/7] Update mismatch migration

itamargiv:
This PR corrects the mismatch migration to fit the most minimal product requirements, keeping in mind length limitations. As it turns out, since we do not intend to do anything with the wikidata value besides showing it, there is no need to save it as a BLOB. In addition, since we limit the length of external values to `1500` then we do not need to store it as TEXT. In case any of these have to be updated in the future, we can create an additional migration to handle that.

### DIFF
--- a/database/migrations/2021_07_19_123858_create_mismatches_table.php
+++ b/database/migrations/2021_07_19_123858_create_mismatches_table.php
@@ -18,8 +18,8 @@ class CreateMismatchesTable extends Migration
             $table->id();
             $table->string('statement_guid');
             $table->string('property_id');
-            $table->binary('wikidata_value'); // Saving as BLOB to enable storing JSON seralizations
-            $table->text('external_value');
+            $table->string('wikidata_value', 2048);
+            $table->string('external_value', 2048);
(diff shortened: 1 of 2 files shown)


database/migrations/2021_07_19_123858_create_mismatches_table.php
```diff
@@ -18,8 +18,8 @@ public function up()
             $table->id();
             $table->string('statement_guid');
             $table->string('property_id');
-            $table->binary('wikidata_value'); // Saving as BLOB to enable storing JSON seralizations
-            $table->text('external_value');
+            $table->string('wikidata_value', 2048);
```
guergana, line 21: Good catch!! I think it's a good idea to add a comment for future maitainers explaining that the 2048 is an arbitrary number above 1500 which is the string length limit that we decided to set for the project.

guergana (changes requested):

database/migrations/2021_07_19_123858_create_mismatches_table.php
itamargiv: :pray: Thank you!

PR #36, opened 2021-07-30
[Imports 1/7] Create CSV reader service

itamargiv:
**Start HERE!** PR **1**/7 in the `feature/import-mismatches` chain.

This change introduces an import CSV reader service to be used in both the mismatch validation and mismatch import jobs. The service lazily reads CSV files from disk, while ensuring that:

1. Column headers match the allowed mismatch column headers
2. Empty rows are skipped
3. Column data is parsed into an associative array according to the specified column headers

In case of issues with column parsing, the service will throw a `ImportParserException`.

Relevant Documentation:
- [Laravel Lazy Collections](https://laravel.com/docs/8.x/collections#lazy-collections)
- [Laravel Exception Context](https://laravel.com/docs/8.x/errors#exception-log-context)

Blogs and Co:
- [Lazy collection CSV Import](https://www.magutti.com/blog/laravel-lazy-collections)

### DIFF
--- a/app/Services/CSVImportReader.php
+++ b/app/Services/CSVImportReader.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace App\Services;
+
+use Illuminate\Support\LazyCollection;
+use App\Exceptions\ImportParserException;
+
+class CSVImportReader
+{
+    public function lines(string $path): LazyCollection
+    {
+        $lines = LazyCollection::make(function () use ($path) {
+            $file = fopen($path, 'r');
+
+            while ($data = fgetcsv($file)) {
+                yield $data;
+            }
+        });
+
+        $allowedKeys = config('imports.upload.column_keys');
+        $columnKeys = $lines->get(0);
+
+        if (count(array_diff($allowedKeys, $columnKeys)) !== 0) {
+            throw new ImportParserException(0, __('parsing.import.headers', [
+                'header-list' => join(", ", $allowedKeys)
+            ]));
+        }
+
+        return $lines->except(0)
+            ->filter(function (array $row) {
+                return $row[0] !== null
+                && $row !== array_fill(0, 5, '');
+            })
+            ->map(function (array $row, int $i) use ($columnKeys) {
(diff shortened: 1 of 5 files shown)


app/Services/CSVImportReader.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace App\Services;
+
+use Illuminate\Support\LazyCollection;
+use App\Exceptions\ImportParserException;
+
+class CSVImportReader
+{
+    public function lines(string $path): LazyCollection
```
Silvan-WMDE, line 10: I would have expected the `$path` as an argument to the constructor, rather than to the `lines()` function.

app/Services/CSVImportReader.php
```diff
@@ -0,0 +1,44 @@
...
+
+            while ($data = fgetcsv($file)) {
+                yield $data;
+            }
+        });
+
+        $allowedKeys = config('imports.upload.column_keys');
+        $columnKeys = $lines->get(0);
+
+        if (count(array_diff($allowedKeys, $columnKeys)) !== 0) {
```
Silvan-WMDE, line 23: this check will not fail if all required header fields are there, but columns are in the wrong order, right? Is that a case we would want to reject?

Silvan-WMDE (commented):
Looks good - two questions below.

app/Services/CSVImportReader.php
itamargiv: I don't think so, because if they are in the wrong order then the file would fail validation, so people will have to make sure that the order of each row matches the order of the columns.

app/Services/CSVImportReader.php
itamargiv: We can do this, but this would mean we will not be able to leverage Laravel's 0 config DI. OR, well we could, but we would have to create a factory class or factory method, which seemed like redundant code to me at that point. What do you think?

Silvan-WMDE (approved):

PR #37, opened 2021-07-30
[Imports 2/7] Create job queue table

itamargiv:
**2**/7 in the `feature/import-mismatches` chain. This depends on #36.

This change introduces a database migration in order to create a `jobs` table so that we could utilize Laravel's database job queue driver.

Relevant documentation:
- [Laravel Job Queue Driver Notes](https://laravel.com/docs/8.x/queues#driver-prerequisites)

Bug: [T285299](https://phabricator.wikimedia.org/T285299)

PR #38, opened 2021-07-30
[Imports 3/7] Create Wikidata API Client

itamargiv:
**3**/7 in the `feature/import-mismatches` chain. This depends on #37.

This change introduces a Wikibase API Client Service, so that we can call Wikidata API endpoints and cache our responses. The service already includes a method to call `wbparsevalue` which throws a `WikibaseValueParserException` on errors. Additionally, we wire up this service to the service container in `WikidataAPIProvider`, as it has external dependencies, and thus cannot be injected with Laravel's Zero-configuration resolution mechanism.

This change also installs some caching middleware, in order to enable us to cache responses for our API calls to try to keep our call rate to a minimum.

Relevant Documentation:
- [Laravel Service Container](https://laravel.com/docs/8.x/container)
- [Laravel Service Providers](https://laravel.com/docs/8.x/providers)
- [Laravel HTTP Client](https://laravel.com/docs/8.x/http-client)
- [Guzzle HTTP Middleware](https://docs.guzzlephp.org/en/stable/handlers-and-middleware.html)
- [Kevinrob's Guzzle Cache Middleware](https://github.com/Kevinrob/guzzle-cache-middleware)

Blogs and Co:
- [When to use DI in Laravel](https://blog.quickadminpanel.com/laravel-when-to-use-dependency-injection-services-and-static-methods/)

### DIFF
--- a/tests/Feature/WikibaseAPIClientTest.php
+++ b/tests/Feature/WikibaseAPIClientTest.php
@@ -74,6 +74,8 @@ class WikibaseAPIClientTest extends TestCase
         $fakeResponseBody = ['test' => 'okay'];
 
         $mockCache = Mockery::mock(CacheMiddleware::class);
+        // The `__invoke()` magic method is utilized by guzzle middleware: 
+        // https://www.phptutorial.net/php-oop/php-__invoke/
         $mockCache->shouldReceive('__invoke')->andReturn(function () use ($fakeResponseBody) {
             return Http::response($fakeResponseBody, 200);
         });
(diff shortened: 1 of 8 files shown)


tests/Feature/WikibaseAPIClientTest.php
```diff
@@ -0,0 +1,100 @@
...
+
+        $this->assertActionRequest(self::FAKE_API_URL, 'wbparsevalue', $fakePayload);
+    }
+
+    public function test_parse_value_retrieves_cached_responses()
+    {
+        $fakeResponseBody = ['test' => 'okay'];
+
+        $mockCache = Mockery::mock(CacheMiddleware::class);
+        $mockCache->shouldReceive('__invoke')->andReturn(function () use ($fakeResponseBody) {
```
itamargiv, line 77:
```suggestion
        // The `__invoke()` magic method is utilized by guzzle middleware: 
        // https://www.phptutorial.net/php-oop/php-__invoke/
        $mockCache->shouldReceive('__invoke')->andReturn(function () use ($fakeResponseBody) {
```

Silvan-WMDE (approved):
LGTM

PR #39, opened 2021-07-30
[Imports 4/7] Create WikidataValue validation Rule

itamargiv:
**4**/7 in the `feature/import-mismatches` chain. This depends on #38.

This change introduces a new custom validation rule for validating wikidata values. The rule utilizes the API client in order to check a Wikibase value against a parser derived from the provided property ID. In case the parse fails, the validation rule will not pass.

Relevant Documentation:
- [Laravel Custom Validation Rules](https://laravel.com/docs/8.x/validation#custom-validation-rules)

### DIFF
--- a/app/Rules/WikidataValue.php
+++ b/app/Rules/WikidataValue.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace App\Rules;
+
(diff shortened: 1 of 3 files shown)


Silvan-WMDE (approved):
looks good, thx!

PR #25, opened 2021-07-23
[Imports 6/7] Create CSV import job

itamargiv:
**Important!** This PR has been re-chained to enable easier review.
**6**/7 in the `feature/import-mismatches` chain. This depends on #24.

In this change, we introduce the CSV import job. The job iterates over a mismatch CSV file and attempts to store each mismatch line into the database. This process is done within a database transaction, to ensure that if an error has occurred while attempting to save, the database will simply roll back to its previous state.

Relevant Documentation:
- [Laravel Database Transactions](https://laravel.com/docs/8.x/database#database-transactions)

Bug: [T285299](https://phabricator.wikimedia.org/T285299)

PR #40, opened 2021-07-30
[Imports 7/7] Dispatch mismatch validation and import on CSV upload

itamargiv:
**END here!** **7**/7 in the `feature/import-mismatches` chain. This depends on #25.

The last change in the mismatches import feature. This change wires up our job dispatching into the `store()` method of our `ImportController`. In order to ensure that the jobs are executed sequentially, we use Laravel's `Bus` facade to queue them together, ensuring that whenever an import file is uploaded, a job will be queued to validate and import that mismatches file into the database.

Relevant Documentation:
- [Laravel Job Chaining](https://laravel.com/docs/8.x/queues#job-chaining)

### DIFF
--- a/app/Http/Controllers/ImportController.php
+++ b/app/Http/Controllers/ImportController.php
@@ -4,10 +4,12 @@ namespace App\Http\Controllers;
 
 use App\Models\ImportMeta;
 use Illuminate\Http\Request;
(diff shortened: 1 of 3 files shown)


Silvan-WMDE (approved):
tested locally - works. Great work, thank you!

PR #41, opened 2021-08-03
Add toolforge background job config

Silvan-WMDE:
This change adds a deployment.yaml file which is needed to create a
kubernetes continuous job on toolforge, that runs the artisan queue
worker to import mismatch files to the database.


itamargiv (approved):
Nice work, this is indeed what we deployed :+1:

PR #45, opened 2021-08-03
Fix UUID Regex from v4 to generic format

itamargiv:
As it turns out, Wikidata can have non v4 UUIDs as the statement GUID. Therefore, we must relax the UUID format rules a bit.

### DIFF
--- a/tests/Feature/ValidateCSVTest.php
+++ b/tests/Feature/ValidateCSVTest.php
@@ -215,6 +215,27 @@ class ValidateCSVTest extends TestCase
         ValidateCSV::dispatch($import);
     }
 
+    public function test_does_not_throw_on_generic_uuid(): void
+    {
+        // Ensure correct columns
+        $line = 'Q4115189$ffa51229-4877-3135-a2e2-a22fe9b7039d' // Emulate non v4 UUID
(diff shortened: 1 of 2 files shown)


tests/Feature/ValidateCSVTest.php
```diff
@@ -215,6 +215,27 @@ public function test_throws_on_invalid_line(Closure $data): void
         ValidateCSV::dispatch($import);
     }
 
+    public function test_does_not_throw_on_generic_uuid(): void
+    {
+        // Ensure correct columns
+        $line = 'Q4115189$ffa51229-4877-3135-a2e2-a22fe9b7039d' // Emulate non v4 UUID
```
Silvan-WMDE, line 221: yay, a real statement guid from the wild outside world... \o/

Silvan-WMDE (approved):
tested locally and works - thanks for the quick fix!

tests/Feature/ValidateCSVTest.php
itamargiv: hehe, also the other ones are super real, they belong to Jane Goodall('s wikidata page)

PR #50, opened 2021-08-05
Create import failure concept

itamargiv:
This change attempts to model the concept of an import failure, which will belong to each failed import. This is done to easily associate import failure messages with their correct import, without having to query the failed jobs table and parsing the serialized exceptions saved there.

To validate, you can attempt to run the migration (`sail artisan migrate`) and the seeder (`sail artisan db:seed ImportFailureSeeder`).

Bug: [T288042](https://phabricator.wikimedia.org/T288042)

### DIFF
--- a/database/factories/ImportFailureFactory.php
+++ b/database/factories/ImportFailureFactory.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace Database\Factories;
+
+use App\Models\ImportFailure;
+use Illuminate\Database\Eloquent\Factories\Factory;
+
+class ImportFailureFactory extends Factory
+{
+    /**
+     * The name of the factory's corresponding model.
+     *
+     * @var string
+     */
+    protected $model = ImportFailure::class;
+
+    /**
+     * Define the model's default state.
+     *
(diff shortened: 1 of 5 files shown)


database/factories/ImportFailureFactory.php
```diff
@@ -0,0 +1,45 @@
...
+            __('validation.import.error', [
+                'line' => $randomLine,
+                'message' => $randomSentence
+            ]),
+            __('parsing.import.error', [
+                'line' => $randomLine,
+                'message' => $randomSentence
+            ])
+        ]) : $randomSentence;
+
```
Silvan-WMDE, line 37: brilliant - I like how much love the details of random message generation get here

Silvan-WMDE (approved):
tested locally - works. :+1:

database/factories/ImportFailureFactory.php
itamargiv: Ah Schtop :blush: Yeah I find it useful to create fakers that try to mimic a situation close to reality, because it helps with the GUI later on.

PR #51, opened 2021-08-06
Persist mismatch failures

itamargiv:
This change utilizes the new `ImportFailure` model to persist import failures for display or in API responses. Additionally, the Job classes were refactored to take advantage of the Job class' `failed()` method and simplify the `InportValidationException` class to remove unnecessary parameters.

Bug: [T288043](https://phabricator.wikimedia.org/T288043)


guergana (approved):
Tested and checked the db locally and works like a charm. Great work!

itamargiv:
Thank you!

PR #52, opened 2021-08-06
Add failure information to import status

itamargiv:
This change adds failure information to both the import status page, and the `GET /api/imports`, `GET /api/imports/{id}` API responses.

Bug: [T288045](https://phabricator.wikimedia.org/T288045)

### DIFF
--- a/app/Models/ImportMeta.php
+++ b/app/Models/ImportMeta.php
@@ -50,4 +50,9 @@ class ImportMeta extends Model
     {
         return $this->belongsTo(User::class)->withDefault();
     }
+
+    public function error()
+    {
+        return $this->hasOne(ImportFailure::class, 'import_id');
+    }
 }
--- a/docs/mismatchfinder-api-spec.yml
+++ b/docs/mismatchfinder-api-spec.yml
@@ -237,4 +237,6 @@ components:
     ListOfImportMeta:
       type: array
       items:
-        $ref: '#/components/schemas/ImportMeta'
+        anyOf:
+          - $ref: '#/components/schemas/ImportMeta'
+          - $ref: '#/components/schemas/FailedImportMeta'
--- a/docs/mismatchfinder-api-spec.yml
+++ b/docs/mismatchfinder-api-spec.yml
@@ -80,6 +80,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ImportMeta'
+
         '400':
           $ref: '#/components/responses/ClientError'
         '422':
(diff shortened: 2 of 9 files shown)


docs/mismatchfinder-api-spec.yml
```diff
@@ -215,9 +224,16 @@ components:
...
 
+    FailedImportMeta:
+      allOf:
+      - $ref: '#/components/schemas/ImportMeta'
+      - type: object
+        properties:
+          error:
+            type: string
+
     ListOfImportMeta:
```
Silvan-WMDE, line 237: this list is returned by `/imports`, and it can now contain instances of both `ImportMeta` and `FailedImportMeta`

docs/mismatchfinder-api-spec.yml
```diff
@@ -101,7 +102,15 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/ImportMeta'
+                oneOf:
+                 - $ref: '#/components/schemas/ImportMeta'
+                 - $ref: '#/components/schemas/FailedImportMeta'
+                discriminator:
```
Silvan-WMDE, line 108: nice. I haven't seen this `discriminator` before

docs/mismatchfinder-api-spec.yml
```diff
@@ -215,9 +224,16 @@ components:
           type: string
           format: date
         uploader:
-          type: object
           $ref: '#/components/schemas/User'
 
+    FailedImportMeta:
+      allOf:
```
Silvan-WMDE, line 230: `allOf:` with an additional property - quick and easy :-)

app/Models/ImportMeta.php
```diff
@@ -50,4 +50,9 @@ public function user()
     {
         return $this->belongsTo(User::class)->withDefault();
     }
+
+    public function error()
+    {
+        return $this->hasOne(ImportFailure::class, 'import_id');
```
Silvan-WMDE, line 56: don't know where else to put this comment, but I noticed that the `ImportMetaSeeder` will still create `ImportMeta` entries with status `failed`, but doesn't seed any `ImportFailure`. This way, both the "Import Status" page and the API endpoint will still show certain imports as "failed", but without an error line.

Silvan-WMDE (changes requested):
Looks good, but a few changes are still needed. I'm actually unsure if they all belong here or if the ImportMetaSeeder one should have been done in PR #51 already.

docs/mismatchfinder-api-spec.yml
itamargiv: :+1:

app/Models/ImportMeta.php
itamargiv: OH! Nice!!! I was actually thinking about it on friday and then completely forgot about it, well caught, thank you!

Silvan-WMDE (approved):
Nice - all fixes work for me. :+1:

PR #61, opened 2021-08-10
Setup vue client side application

itamargiv:


resources/js/Components/Alert.vue
```diff
@@ -0,0 +1,19 @@
+<template>
```
itamargiv, line 1: Not sure if we should commit this, maybe we should include an even simpler component? Something like our website Title or the small svg + log in status link that we have in the blade views?

resources/js/Components/Alert.vue
```diff
@@ -0,0 +1,19 @@
...
+export default {
+    data() {
+        return {
+            message: 'I am an alert.'
+        };
+    }
+};
+</script>
+
+<style>
```
itamargiv, line 15: Perhaps we should try and see how sass works in the style section here, as I'm pretty sure we want to use sass also in our component styles.

resources/js/Pages/SamplePage.vue
```diff
@@ -0,0 +1,23 @@
+<template>
```
itamargiv, line 1: Same as the comment above, I think the example page we include can already be something that we can use as a scaffolding for our app. Maybe we should rename this to be the homepage, and add some "coming soon" copy?

resources/js/app.ts
```diff
@@ -0,0 +1,14 @@
+// import './bootstrap';
+import Vue from 'vue';
+import { createInertiaApp } from '@inertiajs/inertia-vue';
+
+const app = document.getElementById('app');
+
+createInertiaApp({
+  resolve: name => require(`./Pages/${name}`),
```
itamargiv, line 8: Nice and easy!

routes/web.php
```diff
@@ -18,6 +18,8 @@
     return redirect(route('api.settings'));
 });
 
+Route::inertia('/test-inertia', 'SamplePage', [ 'user' => [ 'name' => 'Potato']]);
```
itamargiv, line 21: In the same vain as the comments above, perhaps we should already scaffold the vue app we want to start soon, so we should probably put this in the "/" route (and move the other one to a different route)

tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+        "target": "esnext",
+        "module": "esnext",
+        "allowJs": true,
+        "strict": true,
+        "esModuleInterop": true,
+        "moduleResolution": "node"
+    },
+    "include": [
+        "resources/js",
+        "resources/js/Components"
```
itamargiv, line 12: Not sure if we need this line anymore

resources/js/Pages/SamplePage.vue
```diff
@@ -0,0 +1,23 @@
...
+  <div>
+    <Head title="Welcome" />
```
itamargiv, line 3: Nice example page!

routes/web.php
itamargiv: Also: :potato: :)

itamargiv (commented):
Great work! Thank you :bow: Since I opened this PR, I annot approve or request changes. See comments and changes below

tsconfig.json
guergana: addressed

resources/js/Components/Alert.vue
guergana: yes, sure. sass works out of the box if you add the `uselang` attribute to the tag like this `<style lang="scss">` will add it.

resources/js/Pages/SamplePage.vue
guergana: yes, sure.

routes/web.php
guergana: Will not address this topic in this PR, we will create a separate task for reorganizing the routes as we agreed.

resources/js/Components/Alert.vue
guergana: Addressed. Good idea!

resources/js/Pages/SamplePage.vue
```diff
@@ -0,0 +1,23 @@
...
+    <h1>Mismatch Finder</h1>
+    <p>Welcome, {{ user.name }}. This page is coming soon!</p>
+    <Alert />
```
itamargiv, line 6: I don't think we need this `<Alert />` anymore

routes/web.php
Silvan-WMDE: I understand your hesitation to even touch the existing web routes as part of this PR, but I do think that a basic re-structuring just to outline the general structure would actually be very helpful as part of the vue client side setup. Please find a very simple proposal in #63, which will immediately serve the new client side app from "/", without having to type the test URL in the browser's addres bar.

Silvan-WMDE (dismissed):
It all looks nice and easy and worked out-of-the-box locally. :100:

routes/web.php
Silvan-WMDE: ok, voted down

PR #64, opened 2021-08-11
Setup Wikit

guergana:
* Install Wikit and Wikit-tokens packages.
* Add wikit import to the project's app.scss file
* Add component to test installation
* Replace previous variables with wikit sass token variables.

### DIFF
--- a/resources/js/Pages/Home.vue
+++ b/resources/js/Pages/Home.vue
@@ -4,6 +4,7 @@
     <h1>Mismatch Finder</h1>
     <p>This page is coming soon!</p>
     <auth-widget v-bind:user="user" />
+    <Button type="progressive" variant="primary" @click.native="console">Test Wikit Component</Button>
   </div>
 </template>
 
@@ -11,15 +12,22 @@
     import Vue from 'vue';
     import { Head } from '@inertiajs/inertia-vue'
     import AuthWidget from '../Components/AuthWidget.vue';
+    import { Button } from '@wmde/wikit-vue-components';
 
     export default Vue.extend({
         components: {
             Head,
-            AuthWidget
+            AuthWidget,
+            Button
         },
         props: {
             user: Object,
         },
+        methods: {
+            console() {
+                console.log('testing wikit component');
+            }
+        }
     });
 </script>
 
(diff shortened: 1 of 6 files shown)


resources/sass/app.scss
```diff
@@ -1,5 +1,7 @@
 @import url('https://fonts.googleapis.com/css2?family=Lato&family=Source+Serif+Pro&display=swap');
 
+@import '~@wmde/wikit-vue-components/dist/wikit-vue-components.css';
```
itamargiv, line 3: We probably also want to import the tokens css in here, so that we can use them in the components we create.

resources/js/Pages/Home.vue
```diff
@@ -4,22 +4,30 @@
     <h1>Mismatch Finder</h1>
     <p>This page is coming soon!</p>
     <auth-widget v-bind:user="user" />
+    <Button type="progressive" variant="primary" @click.native="console">Test Wikit Component</Button>
```
itamargiv, line 7: This looks great! What you mean by "adding routing to the button"?

To me, it seems ready: the button looks like we expect it too, and behaves like we expect it too.

itamargiv (changes requested):
One small thing, I think we should also include the wikit tokens package in this case, so that we can use tokens in our "local" components, before we upstream them to wikit.

resources/sass/app.scss
guergana: I thought about it when I read the instructions but didn't know if I should. Good idea! Thanks for pointing it out.

resources/js/Pages/Home.vue
guergana: I wanted to make the button go to the `/store` route. But didn't manage so I removed it. But yes I agree that we can deal with this when we need to.

itamargiv: Ah, in this case we want to use a link that looks like a button, rather than a button.

guergana: Ah, much better then. :)

itamargiv (approved):
Thanks for the change @guergana :100:

Silvan-WMDE (approved):
Tested locally: compiles, loads and works. :+1: 

[T288807](https://phabricator.wikimedia.org/T288807) and [T288809 ](https://phabricator.wikimedia.org/T288809) were created as a result of this review, but still this PR is good to go

PR #66, opened 2021-08-16
Set up eslint for vue/typescript

Silvan-WMDE:
This sets up JavaScript/TypeScript linting via eslint and adds a corresponding GitHub action.


docs/README.md
```diff
@@ -12,6 +12,7 @@
 - [Working with OAuth](#working-with-oauth)
 - [Frontend - Working with CSS and JS](#frontend---working-with-css-and-js)
 - [PHP Linting](#php-linting)
+- [Javascript Linting](#js-linting)
```
itamargiv, line 15: Cool thing, thanks for adding the docs! Do you think these two sections should perhaps live under the same higher level heading "Linting", like such (see below)?

```markdown
## Linting

### PHP Linting

<!-- content... -->

### Javascript Linting

<!-- content... -->
```

package.json
```diff
@@ -1,6 +1,8 @@
 {
     "private": true,
     "scripts": {
+        "lint": "eslint --ext .js,.vue resources/js",
+        "lint:fix": "eslint --fix --ext .js,.vue resources/js",
```
itamargiv, line 5: Nice! :+1: Thank you

itamargiv (commented):
Looks great! I left one comment for discussion, but not sure if it's a blocker :100:

docs/README.md
Silvan-WMDE: Yes, that makes sense. Thank you.

itamargiv (approved):
Thank you!

PR #68, opened 2021-08-16
Get active mismatches

Silvan-WMDE:
This modifies the /mismatches endpoint, so that it only returns mismatches that have not been edited yet and that do not belong to expired imports. Two optional request parameters are added, which allow including reviewed mismatches or expired ones.

Bug: [T285301](https://phabricator.wikimedia.org/T285301)

### DIFF
--- a/app/Http/Controllers/MismatchController.php
+++ b/app/Http/Controllers/MismatchController.php
@@ -3,10 +3,9 @@
 namespace App\Http\Controllers;
 
 use App\Http\Requests\MismatchesRequest;
-use App\Http\Resources\MismatchCollection;
 use App\Http\Resources\MismatchResource;
 use App\Models\Mismatch;
-use Illuminate\Http\Request;
+use Carbon\Carbon;
 
 class MismatchController extends Controller
 {
@@ -20,8 +19,22 @@ class MismatchController extends Controller
      */
     public function index(MismatchesRequest $request)
     {
-        $mismatches = Mismatch::whereIn('item_id', $request->ids)->get();
+        $query = Mismatch::whereIn('item_id', $request->ids);
 
-        return MismatchResource::collection($mismatches);
+        // limit to 'pending',
+        // unless include_reviewed parameter is provided
+        if (!$request->include_reviewed) {
+            $query->where('status', 'pending');
+        }
+
+        // limit to non-expired,
+        // unless include_expired parameter is provided
+        if (!$request->include_expired) {
+            $query->whereHas('importMeta', function ($import) {
+                $import->where('expires', '>=', Carbon::now());
+            });
+        }
+
+        return MismatchResource::collection($query->get());
     }
 }
--- a/app/Http/Controllers/MismatchController.php
+++ b/app/Http/Controllers/MismatchController.php
@@ -5,7 +5,6 @@ namespace App\Http\Controllers;
 use App\Http\Requests\MismatchesRequest;
 use App\Http\Resources\MismatchResource;
 use App\Models\Mismatch;
-use Carbon\Carbon;
 
 class MismatchController extends Controller
 {
@@ -31,7 +30,7 @@ class MismatchController extends Controller
         // unless include_expired parameter is provided
         if (!$request->include_expired) {
             $query->whereHas('importMeta', function ($import) {
-                $import->where('expires', '>=', Carbon::now());
+                $import->where('expires', '>=', now());
             });
         }
 
--- a/database/factories/MismatchFactory.php
+++ b/database/factories/MismatchFactory.php
@@ -31,7 +31,7 @@ class MismatchFactory extends Factory
     }
 
     /**
-     * Indicate that a mismatch has been edited
+     * Indicate that a mismatch has been reviewed
      *
      * @return \Illuminate\Database\Eloquent\Factories\Factory
      */
--- a/database/factories/MismatchFactory.php
+++ b/database/factories/MismatchFactory.php
@@ -65,8 +65,11 @@ class MismatchFactory extends Factory
 
     private function getRandomReviewStatus()
     {
-        $reviewStatuses = [ 'wikidata', 'external', 'both', 'none' ];
-        $randomStatusIndex = $this->faker->numberBetween(0, 3);
-        return $reviewStatuses[ $randomStatusIndex ];
+        return $this->faker->randomElement([
+            'wikidata',
+            'external',
+            'both',
+            'none'
+        ]);
     }
 }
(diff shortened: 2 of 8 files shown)


app/Http/Controllers/MismatchController.php
```diff
@@ -20,8 +19,22 @@ class MismatchController extends Controller
...
+        // unless include_reviewed parameter is provided
+        if (!$request->include_reviewed) {
+            $query->where('status', 'pending');
+        }
+
+        // limit to non-expired,
+        // unless include_expired parameter is provided
+        if (!$request->include_expired) {
+            $query->whereHas('importMeta', function ($import) {
+                $import->where('expires', '>=', Carbon::now());
```
itamargiv, line 34:
```suggestion
                $import->where('expires', '>=', now());
```

I don't think we actually need to use Carbon directly here, as carbon is not a direct dependency of our app and  Laravel provides a [global `now()` helper](https://laravel.com/docs/8.x/helpers#method-now) for this use case. This will also help with ["stopping time" in tests](https://laravel.com/docs/8.x/mocking#interacting-with-time) (See #15 for a usage example in our repo).

app/Http/Controllers/MismatchController.php
```diff
@@ -3,10 +3,9 @@
 namespace App\Http\Controllers;
 
 use App\Http\Requests\MismatchesRequest;
-use App\Http\Resources\MismatchCollection;
 use App\Http\Resources\MismatchResource;
 use App\Models\Mismatch;
-use Illuminate\Http\Request;
+use Carbon\Carbon;
```
itamargiv, line 8: Same as above/below

database/factories/MismatchFactory.php
```diff
@@ -48,4 +62,11 @@ private function getRandomValue()
             $this->faker->words($randomWordAmount, true)
         ]);
     }
+
+    private function getRandomReviewStatus()
+    {
+        $reviewStatuses = [ 'wikidata', 'external', 'both', 'none' ];
+        $randomStatusIndex = $this->faker->numberBetween(0, 3);
+        return $reviewStatuses[ $randomStatusIndex ];
```
itamargiv, line 70: Nice! We can also use [faker's  `randomElement()`](https://fakerphp.github.io/formatters/numbers-and-strings/#randomelement) for this purpose:

```suggestion
        return $this->faker->randomElement([
            'wikidata',
            'external',
            'both',
            'none'
        ]);
```

database/factories/MismatchFactory.php
```diff
@@ -30,6 +30,20 @@ public function definition()
         ];
     }
 
+    /**
+     * Indicate that a mismatch has been reviewed
+     *
+     * @return \Illuminate\Database\Eloquent\Factories\Factory
+     */
+    public function reviewed()
```
itamargiv, line 38: I really like the use of states we have in our factories :+1:

database/seeders/MismatchSeeder.php
```diff
@@ -20,8 +20,27 @@ public function run()
...
+            ->create();
+
+        Mismatch::factory(10)
             ->for($import)
+            ->reviewed()
+            ->create();
+
+        Mismatch::factory(11)
+            ->for($expiredImport)
+            ->reviewed()
```
itamargiv, line 43: 10+11+10+11 = 42 Checks out :nerd_face: :heart_eyes:

docs/mismatchfinder-api-spec.yml
```diff
@@ -195,6 +195,14 @@ components:
           type: string
         external_url:
           type: string
+        status:
+          type: string
+          enum:
```
itamargiv, line 200: :100: Thanks for updating the specs!

tests/Feature/ApiMismatchRouteTest.php
```diff
@@ -62,21 +62,133 @@ public function test_single_mismatch_returns_correct_data()
...
-    public function test_multiple_mismatches_returns_correct_number_of_items()
+    public function test_multiple_mismatches_do_not_return_edited_or_expired()
     {
         $import = ImportMeta::factory()
         ->for(User::factory()->uploader())
         ->create();
 
-        $mismatches = Mismatch::factory(3)->for($import)->create();
+        $expiredImport = ImportMeta::factory()
+        ->for(User::factory()->uploader())
```
itamargiv, line 72: Totally a nitpick, but I wonder if we should add a phpcs rule for indentation levels. Visually it bugs me, but as long as the code isn't broken, it's not a hill I'm prepared to die over.

tests/Feature/ApiMismatchRouteTest.php
```diff
@@ -62,21 +62,133 @@ public function test_single_mismatch_returns_correct_data()
...
+        $expiredMismatches = Mismatch::factory(3)->for($expiredImport)->create();
+        $expiredReviewedMismatches = Mismatch::factory(3)->for($expiredImport)->reviewed()->create();
+        $response = $this->json(
+            'GET',
+            self::MISMATCH_ROUTE,
+            [
+                'ids' => $pendingMismatches->implode('item_id', '|') . '|' .
+                         $reviewedMismatches->implode('item_id', '|') . '|' .
+                         $expiredMismatches->implode('item_id', '|') . '|' .
+                         $expiredReviewedMismatches->implode('item_id', '|')
```
itamargiv, line 87: Well constructed!

itamargiv (changes requested):
Looks great! Thank you, :bow: Just one thing to change and a couple of suggestions

database/factories/MismatchFactory.php
Silvan-WMDE: ah, that's better. thx.

tests/Feature/ApiMismatchRouteTest.php
Silvan-WMDE: I couldn't find a phpcs rule that checks for the correct indentation level in this case. PSR-2 only defines that there must be at least 8 spaces, but it would also be happy with 13. I stopped searching the internet after 30 minutes.

Still fixed the indentation, thx.

database/seeders/MismatchSeeder.php
```diff
@@ -20,8 +20,27 @@ public function run()
             ->for(User::factory()->uploader())
             ->create();
 
-        Mismatch::factory(42)
+        $expiredImport = ImportMeta::factory()
+            ->for(User::factory()->uploader())
+            ->expired()
+            ->create();
```
itamargiv, line 26: We probably also want to make sure that the status of the expired import is 'completed'. Otherwise, we might get mismatches in the seeded database that come from pending of failed imports.

```suggestion
            ->create([
                'status' => 'completed'
            ]);
```

itamargiv (changes requested):
Perfect, thank you for all the changes! One last tiny comment and we're good to go!

itamargiv (approved):

itamargiv:
Thank you!

PR #69, opened 2021-08-18
Re-structure web routes for use with vue app

itamargiv:
As we intend to serve our client side application from our app's main route, a few changes have to be made to our app's preliminary routing model. This change ensures that we are ready to serve the application from the website's root.

Bug: [T289251](https://phabricator.wikimedia.org/T289251)


tests/Feature/WebRouteTest.php
```diff
@@ -5,33 +5,83 @@
...
+     */
+    public function test_store_route()
+    {
+        $response = $this->get('/store');
+
+        $response->assertRedirect(route('store.api-settings'));
+    }
+
+    /**
+     * Test the authenticated /store/api-settings route
```
Silvan-WMDE, line 63: Hm, actually, this test now covers both, authenticated and non-authenticated access to that route. Does that mean browser tests are needed to actually test the difference between the two?

PS: Found it.

tests/Browser/ApiSettingsTest.php
```diff
@@ -5,9 +5,9 @@
...
 use Laravel\Dusk\Browser;
 use Tests\DuskTestCase;
-use Tests\Browser\Pages\HomePage;
+use Tests\Browser\Pages\StorePage;
 
-class ExampleTest extends DuskTestCase
+class ApiSettingsTest extends DuskTestCase
 {
     /**
      * A basic browser test example.
```
Silvan-WMDE, line 13: Ah, here's the browser test that I was missing in one of my comments below. Maybe this is not so basic anymore, since it actually tests the un-authenticated view? And maybe it's worth adding a second one for the authenticated user, with the expecation to see a "Log out" link? Or is that too complicated to manage in a browser test?

tests/Browser/Pages/StorePage.php
```diff
@@ -13,7 +13,7 @@ class HomePage extends Page
      */
     public function url()
     {
-        return '/';
+        return '/store/api-settings/';
```
Silvan-WMDE, line 16: What is this `url()` needed for? And can we replace it with `route('store.home')` so that the URL path isn't defined in multiple places?

tests/Feature/AuthTokenTest.php
```diff
@@ -71,7 +43,7 @@ public function test_createTokenWhenNoneExists_returnsNewToken()
                          ->get('/auth/create-token');
 
         $response->assertStatus(302);
-        $response->assertRedirect('auth/api-settings');
+        $response->assertRedirect('store/api-settings');
```
Silvan-WMDE, line 46: again: can we use `route('store.api-settings')` here, instead?

Silvan-WMDE (changes requested):
Ok, so this is big, and tbh I don't think it should even be considered a part of [T287356](https://phabricator.wikimedia.org/T287356). I think we should have made that a separate tech tasks, that needs to be done before any of the app UI can be touched.

Anyways: lots of comments. More questions than answers, without any strong opinion on my side. Let's just agree on what are the best approaches.

tests/Feature/WebRouteTest.php
```diff
@@ -5,33 +5,83 @@
...
     /**
-     * Test the /auth/api_settings route
+     * Test the / route
      *
      *  @return void
      */
-    public function test_welcome_route()
+    public function test_home_route()
     {
         $response = $this->get('/');
```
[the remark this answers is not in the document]
itamargiv: I like thais, and I'm glad you suggested it :smile:

tests/Browser/ApiSettingsTest.php
itamargiv: yeah, these are the browser tests I was referring to yesterday, see #70 for more info

tests/Browser/Pages/StorePage.php
itamargiv: I think it didn't work when I tried it, but I will try again :smile:

itamargiv: P.S. The url is needed in the tests, when you `visit(new StorePage)`

tests/Browser/ApiSettingsTest.php
itamargiv, line 13:
```suggestion
     * Guests can see log in link
```

tests/Browser/ApiSettingsTest.php
```diff
@@ -17,7 +17,7 @@ class ExampleTest extends DuskTestCase
     public function testBasicExample()
```
itamargiv, line 17:
```suggestion
    public function testGuestSeesLoginLink()
```

tests/Browser/Pages/StorePage.php
itamargiv:
```suggestion
        return route('store.home');
```

tests/Feature/AuthTokenTest.php
itamargiv:
```suggestion
        $response->assertRedirect(route('store.api-settings'));
```

tests/Feature/WebRouteTest.php
```diff
@@ -5,33 +5,83 @@
...
+                    ->where('user.name', $user->username);
+            });
+    }
+
+    /**
+     * Test the /store route
+     *
+     *  @return void
+     */
+    public function test_store_route()
```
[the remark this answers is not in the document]
itamargiv:
```suggestion
    public function test_store_home()
```

tests/Feature/WebRouteTest.php
```diff
@@ -5,33 +5,83 @@
...
+
+        $response->assertRedirect(route('store.api-settings'));
+    }
+
+    /**
+     * Test the authenticated /store/api-settings route
+     *
+     *  @return void
+     */
+    public function test_api_settings_returnsShowToken()
```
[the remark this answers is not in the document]
itamargiv:
```suggestion
    public function test_store_api_settings()
```

tests/Feature/WebRouteTest.php
```diff
@@ -5,33 +5,83 @@
...
+        $response->assertViewIs('showToken');
     }
 
     /**
-     * Test the /importStatus route
+     * Test the store/imports route
      *
      *  @return void
      */
     public function test_importStatus_route()
```
itamargiv, line 80:
```suggestion
    public function test_store_import_status()
```

Silvan-WMDE (approved):
LGTM - Thanks for including all the suggested changes. :100:

PR #70, opened 2021-08-18
Add more browser tests

itamargiv:
These browser tests were originally included in PR #69. However, since they were only failing in CI, and more investigation is needed, these were moved out to this change in order to unblock #69.

Bug: [T289154](https://phabricator.wikimedia.org/T289154)


itamargiv:
No longer Relevant after #70

PR #71, opened 2021-08-19
Set up localization

itamargiv:
This change sets up localization for our client side application, by adding a middleware to infer the locale from a `uselang` query parameter, and requesting the correct translation strings from the public folder using a small HTTP client. The translation strings should follow the banana-i18n format, as is common in other WMDE tools.

Additionally, a provisional Error page component has been added, in case an error occurs while trying to retrieve localization files or starting up the client-side application.

Bug: [T287356](https://phabricator.wikimedia.org/T287356)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -165,7 +165,13 @@ sail npm run dev
 
 ## Localization and Internationalization
 
-The Mismatch Finder application employs two separate localization systems: one for the server-side Laravel  app, and another for the vue based clinet-side application.
+The Mismatch Finder application employs two separate localization systems: one for the server-side Laravel  app, and another for the vue based client-side application.
+
+To switch to any other language than English, set the `uselang` parameter in the URL. For example, to set the language to German:
+
+```
+http://<your-localhost>/?uselang=de
+```
 
 On the server side, we fully employ the default [Laravel localization system](https://laravel.com/docs/8.x/localization) and syntax, and messages in the server side should be added as or to php files in the `resources/lang/en` directory.
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -163,6 +163,20 @@ To manually compile assets for your local dev environment run:
 sail npm run dev
 ```
 
+## Localization and Internationalization
+
+The Mismatch Finder application employs two separate localization systems: one for the server-side Laravel  app, and another for the vue based clinet-side application.
+
+On the server side, we fully employ the default [Laravel localization system](https://laravel.com/docs/8.x/localization) and syntax, and messages in the server side should be added as or to php files in the `resources/lang/en` directory.
+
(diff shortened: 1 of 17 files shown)


docs/README.md
```diff
@@ -163,6 +163,20 @@ To manually compile assets for your local dev environment run:
 sail npm run dev
 ```
 
+## Localization and Internationalization
+
+The Mismatch Finder application employs two separate localization systems: one for the server-side Laravel  app, and another for the vue based clinet-side application.
+
+On the server side, we fully employ the default [Laravel localization system](https://laravel.com/docs/8.x/localization) and syntax, and messages in the server side should be added as or to php files in the `resources/lang/en` directory.
+
+The client side localization system utilizes the [banana-i18n](https://www.npmjs.com/package/banana-i18n) library and format. Messages for the client side application are kept in the `public/i18n/` directory, and are served in order to be consumed by the vue client. Each message that is added to the `en.json` file, should be documented in the `qqq.json` file as well, to provide more context for translators.
```
guergana, line 172: Good docs. Thanks!

tests/Feature/WebRouteTest.php
```diff
@@ -47,6 +49,34 @@ public function test_home_page_has_user()
...
+        $response->assertSuccessful();
+        $this->assertSame(App::currentLocale(), 'qqq');
+    }
+
+    /**
+     * Test the / route with a non translated language code
+     *
+     *  @return void
+     */
+    public function test_home_page_locale_falls_back_to_english()
```
guergana, line 71: Nice add!!!

docs/README.md
guergana: Maybe we should mention in the docs that we can pass the  `uselang=LANGUAGE` in the route to change to other valid languages?

guergana (changes requested):
Great!!! I would add a comment mentioning the `uselang` parameter to test in the docs. I don't know if everyone is familiar with this.

docs/README.md
itamargiv: Great Idea! Thank you!

itamargiv: Done :dancing_men:

Silvan-WMDE (commented):
This all looks good to me and it is working locally, but I'll leave it to @guergana to approve, as she did a thorough review and had change requests. Thanks for setting this up!

docs/README.md
```diff
@@ -165,7 +165,13 @@ sail npm run dev
 
 ## Localization and Internationalization
 
-The Mismatch Finder application employs two separate localization systems: one for the server-side Laravel  app, and another for the vue based clinet-side application.
+The Mismatch Finder application employs two separate localization systems: one for the server-side Laravel  app, and another for the vue based client-side application.
+
+To switch to any other language than English, set the `uselang` parameter in the URL. For example, to set the language to German:
```
guergana, line 170: Thanks for adding this.

guergana (approved):

PR #72, opened 2021-08-20
Fix buggy check of request parameters

Silvan-WMDE:
This fix now properly checks whether the request parameters
'include_reviewed' and 'include_expired' have "truthy" values, even if
they are actually strings such as 'true', 'yes' or 'on'. Other values,
including string 'false', are considered to be not true.

Additional feature tests ensure the correct behaviour.

Bug: T285301

### DIFF
--- a/app/Http/Controllers/MismatchController.php
+++ b/app/Http/Controllers/MismatchController.php
@@ -22,13 +22,13 @@ class MismatchController extends Controller
 
         // limit to 'pending',
         // unless include_reviewed parameter is provided
-        if (!$request->include_reviewed) {
+        if (!$request->boolean('include_reviewed')) {
             $query->where('status', 'pending');
         }
 
         // limit to non-expired,
         // unless include_expired parameter is provided
-        if (!$request->include_expired) {
+        if (!$request->boolean('include_expired')) {
             $query->whereHas('importMeta', function ($import) {
                 $import->where('expires', '>=', now());
             });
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/MismatchController.php
```diff
@@ -22,13 +22,13 @@ public function index(MismatchesRequest $request)
 
         // limit to 'pending',
         // unless include_reviewed parameter is provided
-        if (!$request->include_reviewed) {
+        if ($request->include_reviewed != 'true') {
```
itamargiv, line 25: oh my, I can't believe I missed that :man_facepalming:

tests/Feature/ApiMismatchRouteTest.php
```diff
@@ -103,6 +103,25 @@ public function test_query_including_reviewed_returns_reviewed_mismatches()
             );
     }
 
+    public function test_query_including_reviewed_false_does_not_return_reviewed_mismatches()
```
itamargiv, line 106: :+1: thanks for the added test!

itamargiv (approved):
Looks good, passes and works. Thank you!

app/Http/Controllers/MismatchController.php
itamargiv: Non blocking, but this looks like another - more robust way of handling boolean query params: https://laravel.com/docs/8.x/requests#retrieving-boolean-input-values

itamargiv (approved):
Fantastic, thank you! :100:

PR #74, opened 2021-08-20
Add textarea component

itamargiv:
This PR creates the textarea components with the minimum specs required for the Mismatch Finder. Disabled / Warning, Error states and porting to Wikit will be added in subsequent PRs.

Bug: [T287568](https://phabricator.wikimedia.org/T287568)

### DIFF
--- a/resources/js/Components/TextArea.vue
+++ b/resources/js/Components/TextArea.vue
@@ -13,7 +13,7 @@
         <textarea :class="[
             'wikit-TextArea__textarea',
             `wikit-TextArea__textarea--${resizeType}`
-        ]" :rows="rows" :placeholder="placeholder" label=""></textarea>
+        ]" :rows="rows" :placeholder="placeholder" label="" @input="$emit( 'input', $event.target.value )"></textarea>
     </div>
 </template>
 
--- a/resources/js/Components/TextArea.vue
+++ b/resources/js/Components/TextArea.vue
@@ -109,7 +109,6 @@ export default Vue.extend({
         color: $wikit-Input-color;
         background-color:  $wikit-Input-background-color;
 
-
         /**
          * Typography
          */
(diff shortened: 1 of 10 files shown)


tests/Vue/Components/TextArea.spec.js
```diff
@@ -22,6 +22,14 @@ describe('TextArea.vue', () => {
         expect(wrapper.find('textarea').classes()).toContain('wikit-TextArea__textarea--horizontal');
     });
 
+    it('ignores invalid resize values', () => {
+        const wrapper = mount(TextArea, {
+            propsData: { resize: 'nonsense' }
+        });
+
+        expect(wrapper.find('textarea').classes()).not.toContain('wikit-TextArea__textarea--nonsense');
```
guergana, line 30: Good workaround! :D and much simpler

resources/js/Components/TextArea.vue
```diff
@@ -102,28 +102,55 @@ export default Vue.extend({
...
-        background-color:  $wikit-Input-background-color;
-        color: $wikit-Input-color;
-
-        font-family: $wikit-Input-font-family;
-        font-size: $wikit-Input-font-size;
-        font-weight: $wikit-Input-font-weight;
-        line-height: $wikit-Input-line-height;
-
+        /**
+         * Animation
```
guergana, line 140: nice and fancy.

resources/js/Components/TextArea.vue
```diff
@@ -56,7 +56,7 @@ export default Vue.extend({
         resizeType: function(){
             // Unfortunately, the vue prop validator does not throw or falls
             // back to default values on validation failure, therefore, we need
-            // to check for a valid
+            // to check for a valid resize limit value
```
guergana, line 59: heheh, so exact

tests/Vue/Components/TextArea.spec.js
Silvan-WMDE: I think I disagree and this should be:
```suggestion
        expect(wrapper.find('textarea').classes()).toContain('wikit-TextArea__textarea--vertical');
```
because that's what we put in the class list for textarea for empty or nonsense `resize` values.

Silvan-WMDE (changes requested):

itamargiv:
Nice catch, thank you!

Silvan-WMDE (approved):
Ok, I'm certainly not the best person to approve this, but it works. And it does what the ticket says. And I found it to be quite usable when playing around locally. So let's get this merged. :100:

PR #75, opened 2021-08-23
Update the Mismatch concept

Silvan-WMDE:
This modifies the Mismatch model to include a reviewer and renames the
'status' field to 'review_status

For these changes to take effect, several components had to be touched:
 * the Mismatch model,
 * migrations for the 'mismatches' database table,
 * the factory, controller and resource,
 * the API route test.

Since Laravel currently does not support renaming an 'enum' column, the
db migration drops and re-creates it at the same position with 'pending'
as its default value. This will do no harm, because no review decisions
have been recorded so far. SQLite does not even allow dropping and
adding columns, therefore in the 'testing' environment, the entire
table will be dropped and re-created with the new column name.

Bug: [T289132](https://phabricator.wikimedia.org/T289132)

### DIFF
--- a/app/Http/Controllers/MismatchController.php
+++ b/app/Http/Controllers/MismatchController.php
@@ -23,7 +23,7 @@ class MismatchController extends Controller
         // limit to 'pending',
         // unless include_reviewed parameter is provided
         if (!$request->boolean('include_reviewed')) {
-            $query->where('status', 'pending');
+            $query->where('review_status', 'pending');
         }
 
         // limit to non-expired,
--- a/app/Http/Resources/MismatchResource.php
+++ b/app/Http/Resources/MismatchResource.php
@@ -22,7 +22,8 @@ class MismatchResource extends JsonResource
             'wikidata_value' => $this->wikidata_value,
             'external_value' => $this->external_value,
             'external_url' => $this->external_url,
-            'status' => $this->status,
+            'review_status' => $this->review_status,
+            'reviewer' => new UserResource($this->user),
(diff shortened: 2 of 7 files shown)


app/Http/Resources/MismatchResource.php
```diff
@@ -22,7 +22,8 @@ public function toArray($request)
             'wikidata_value' => $this->wikidata_value,
             'external_value' => $this->external_value,
             'external_url' => $this->external_url,
-            'status' => $this->status,
+            'review_decision' => $this->review_decision,
+            'reviewer' => new UserResource($this->user),
```
guergana, line 26: good idea. to add the reviewer as well

app/Http/Controllers/MismatchController.php
```diff
@@ -23,7 +23,7 @@ public function index(MismatchesRequest $request)
         // limit to 'pending',
         // unless include_reviewed parameter is provided
         if (!$request->boolean('include_reviewed')) {
-            $query->where('status', 'pending');
+            $query->where('review_decision', 'pending');
```
guergana, line 26: was it decided to change it to `review_decision`. if my memory doesn't fail i remember we said `review_status` sounded better.

guergana (approved):
LGTM!! :100:

app/Http/Controllers/MismatchController.php
Silvan-WMDE: Yes, I think you're right. Changed.

itamargiv (approved):
Looks good to me!

PR #76, opened 2021-08-24
Align class name with file name

Silvan-WMDE:
This quick fix allows developers to run
`sail artisan test tests/Feature/ApiUserRouteTest.php`
which previously failed because:

> Class 'ApiUserRouteTest' could not be found


itamargiv (approved):
Very well, :smile: thank you!

PR #78, opened 2021-08-25
suggested change to update mismatches

Silvan-WMDE:
it was easier to upload my suggested changes in a branch than to put them all in a review


app/Http/Controllers/MismatchController.php
```diff
@@ -46,14 +44,11 @@ public function index(MismatchesRequest $request)
...
-        $mismatch = Mismatch::find($id);
-
         // TODO: validate value of review_status
         // TODO: changes to any other property will lead to a validation error
-        
-        $mismatch->review_status = $request->review_status;
-        $mismatch->save();
-       
+        $mismatch = Mismatch::find($id);
+        $mismatch->update($request->all());
```
guergana, line 50: I don't think this is what we want to do. We don't want to be able to modify all the parameters.

guergana (changes requested):
I think that we don't need the `$mismatch->update($request->all());` 
The rest are very good changes. Thanks for adding that.

PR #82, opened 2021-08-27
Add querying section to home page

itamargiv:
This change adds the querying section to the home page of the Mismatch Finder application. This section contains a form to send item IDs to the server, which will be used to filter mismatches to retrieve. The submit button is currently disabled, as we cannot yet display the results of the form.

Bug: [T289057](https://phabricator.wikimedia.org/T289057)


guergana (approved):
Looks awesum! Will need a second check from UX but looks awesome nonetheless

PR #83, opened 2021-08-31
Add Mismatches put request to api specs

itamargiv:
Bug: [T289133](https://phabricator.wikimedia.org/T289133)

### DIFF
--- a/docs/mismatchfinder-api-spec.yml
+++ b/docs/mismatchfinder-api-spec.yml
@@ -232,6 +232,9 @@ components:
             updated_at:
               type: string
               format: date-time
+            reviewer:
+              type: object
+              $ref: '#/components/schemas/User'
         - $ref: '#/components/schemas/FillableMismatch'
 
     ListOfMismatches:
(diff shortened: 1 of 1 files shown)


docs/mismatchfinder-api-spec.yml
```diff
@@ -38,6 +38,24 @@ paths:
         '422':
           $ref: '#/components/responses/ValidationErrors'
 
+    put:
```
Silvan-WMDE, line 41: shouldn't this have the {id} parameter in the URL?

```
  /mismatches/{id}:
    put:
      tags: ['store']
      summary: Update Mismatch review status
      parameters:
        - in: path
          name: id
          schema:
            type: number
          required: true

```

Silvan-WMDE (changes requested):
Thanks for spotting the missing spec!

docs/mismatchfinder-api-spec.yml
itamargiv: :facepalm: It for sure should...

Silvan-WMDE (approved):
nice, thanks!

PR #84, opened 2021-08-31
Write review log messages to dedicated file

Silvan-WMDE:
This change moves the log message upon update of a mismatch's
review_status to the 'mismatch_updates.log' file.

A test to ensure that the correct message is logged to the correct log
channel makes use of the 'timacdonald/log-fake' composer package.

Bug: [T289134](https://phabricator.wikimedia.org/T289134)


itamargiv (approved):
Looks Good To Me :100:

PR #85, opened 2021-08-31
Adjust maximal body width for Mismatch Finder SPA

itamargiv:
This change adjusts the maximal body width of the Mismatch Finder Website, to match design specifications.

Bug: [T294565](https://phabricator.wikimedia.org/T294565)

### DIFF
--- a/resources/sass/app.scss
+++ b/resources/sass/app.scss
@@ -9,10 +9,13 @@
 * Layout
 */
 body {
-    max-width: 1090px;
     padding: 0 $dimension-layout-xsmall;
     margin: 0 auto;
     @include body-text;
+
+    &.app-container {
+        max-width: calc(1132px + (#{$dimension-layout-xsmall} * 2)); // Compensate for extra padding in large screens
+    }
 
     @media(min-width: $width-breakpoint-tablet){
         padding: 0 $dimension-layout-small;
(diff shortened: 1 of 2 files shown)


itamargiv:
@SaiSan-WMDE @MattiaCapozziWMDE any news on this?

SaiSan-WMDE:
> @SaiSan-WMDE @MattiaCapozziWMDE any news on this?

Hey, Itamar! Mattia and I had a couple of sessions to try to find a "soft-spot" value for the max-width of the content (a size that will allow displaying all the information in our quite dense table, and yet doesn't stretch the width of the text area in the main page too much). We found that value to be 1168px (including margins) or 73em. In principle, a new ticket will be created for this issue.

resources/sass/app.scss
```diff
@@ -9,10 +9,10 @@
 * Layout
 */
 body {
-    max-width: 1090px;
     padding: 0 $dimension-layout-xsmall;
     margin: 0 auto;
     @include body-text;
+    max-width: 1168px;
```
Silvan-WMDE, line 15: any particular reason why you changed it from first to fourth row? :smile:

Silvan-WMDE (dismissed):
Yes, this makes the most sense to me, if the width is always 1168px, which already includes the padding. No calculations needed here.

But let's wait what @itamargiv thinks, just before before merging it.

itamargiv (commented):
I cannot request changes, since this is my PR, but this change should only affect the inertia app and not the "store" pages, therefore the app container should not have been removed.

Silvan-WMDE (approved):
✅️

PR #88, opened 2021-09-02
Use Toolforge FontCDN proxy

supertassu:
Please see https://wikitech.wikimedia.org/wiki/Help:Toolforge/Web#Load_external_assets_using_our_CDN_services


Silvan-WMDE (approved):
Thanks for the suggestion!

PR #90, opened 2021-09-03
Create endpoint for TextArea input of Item IDs 

Silvan-WMDE:
Bug: [T290155](https://phabricator.wikimedia.org/T290155)


Silvan-WMDE:
covered by #89

PR #91, opened 2021-09-03
Send item IDs to the server

Silvan-WMDE:
This change makes use of the '/results' endpoint and Inertia response created in #90. It parses item IDs from the TextArea, sends them to the server and displays the simple Results page.

Bug: [T290158](https://phabricator.wikimedia.org/T290158)

### DIFF
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -2,9 +2,9 @@
   <div>
     <Head title="Mismatch Finder - Results" />
     <h1>Mismatch Finder - Results!</h1>
-    <div class="results">
+    <p class="results">
         Thank you for sending IDs {{item_ids}}. Mismatch results will be displayed in this page soon.
-    </div>
+    </p>
   </div>
 </template>
 
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -2,7 +2,9 @@
   <div>
     <Head title="Mismatch Finder - Results" />
     <h1>Mismatch Finder - Results!</h1>
-    <p>Thank you for sending IDs {{item_ids}}. Mismatch results will be displayed in this page soon.</p>
+    <div class="results">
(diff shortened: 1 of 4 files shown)


resources/js/Pages/Results.vue
```diff
@@ -2,7 +2,9 @@
   <div>
     <Head title="Mismatch Finder - Results" />
     <h1>Mismatch Finder - Results!</h1>
-    <p>Thank you for sending IDs {{item_ids}}. Mismatch results will be displayed in this page soon.</p>
+    <div class="results">
```
itamargiv, line 5: I think this can still stay a `p` with the class results

tests/Browser/ItemsFormTest.php
```diff
@@ -22,4 +22,16 @@ public function test_can_enter_list_of_item_ids()
                     ->assertInputValue('@items-input', "Q1\nQ2");
         });
     }
+
+    public function test_can_submit_list_of_item_ids()
```
itamargiv, line 26: Nice addition for the test! Thank you!

tests/Browser/ItemsFormTest.php
```diff
@@ -22,4 +22,16 @@ public function test_can_enter_list_of_item_ids()
...
+    {
+        $this->browse(function (Browser $browser) {
+            $browser->visit(new HomePage)
+                    ->keys('@items-input', 'Q1', '{return_key}', 'Q2')
+                    ->press('button')
+                    ->waitFor('div.results')
```
itamargiv, line 32: Instead of referring to the classes directly, I think It's best to add shortcuts for them in the `elements` method of the Page object we use. That way, if anything changes in the client side app, we will only need to update it in one place.

itamargiv (changes requested):
Great work! Just one small change and we are good to go!

itamargiv (approved):
Looks great, works great, thank you!

PR #92, opened 2021-09-03
Add Website header

guergana:
Bug: [T287571](https://phabricator.wikimedia.org/T287571)


public/images/user.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="14" height="16" viewBox="0 0 14 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Silvan-WMDE, line 1: this and the following image `public/images/wikidata-logo.svg` are generated, right? So I guess they shouldn't be added to git.

resources/js/Components/AuthWidget.vue
```diff
@@ -1,11 +1,13 @@
...
-    </div>
-    <div v-else class="guest">
-        <!-- Stuff to Show otherwise -->
-        <p>Hello, Guest!</p>
+    <div>
+        <div v-if="user">
+            <a :href="`https://www.wikidata.org/wiki/User:${user.name}`">
+                <img src="../../img/user.svg" class="icon-user" /><span class="username">{{ user.name }}</span>
+                </a>
+            <a href="/auth/logout">Logout</a>
```
Silvan-WMDE, line 7: Shouldn't these two (see Log in below) be i18n'd? Also: "Log out", as per the figma spec.

Silvan-WMDE (changes requested):
Looks great - thanks! :100: 
Two minor changes below and maybe the new Layout is also worth a quick browser test?

webpack.mix.js
```diff
@@ -15,3 +15,21 @@ mix.ts('resources/js/app.ts', 'public/js')
     .vue({ version: 2 })
     .sass('resources/sass/app.scss', 'public/css');
 
+mix.override(webpackConfig => {
+    // BUG: vue-loader doesn't handle file-loader's default esModule:true setting properly causing
+    // <img src="[object module]" /> to be output from vue templates.
```
itamargiv, line 20: I am not sure what this is for? You aren't importing the image object itself to be used in the source, you are using the url that is generated. Seems to me that file copying from resources to public happens by default: https://laravel.com/docs/8.x/mix#url-processing

resources/js/Components/AuthWidget.vue
```diff
@@ -1,11 +1,13 @@
...
-    <div v-if="user" class="auth">
-        <!-- Stuff to show when user is logged in -->
-        <p>Hello, {{user.name}}!</p>
-    </div>
-    <div v-else class="guest">
-        <!-- Stuff to Show otherwise -->
-        <p>Hello, Guest!</p>
+    <div>
+        <div v-if="user">
+            <a :href="`https://www.wikidata.org/wiki/User:${user.name}`">
```
itamargiv, line 4: I wonder if this link should open in a new tab?

itamargiv (changes requested):
Fantastic work! Thank you very much :pray: I left a couple of comments about the svgs, but apart from that, lookig great!

resources/js/Pages/Layout.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+  <main>
+    <header>
+      <img src="../../img/wikidata-logo.svg" alt="Wikidata-logo" />
```
itamargiv, line 4: Same as above
```suggestion
      <img src="images/wikidata-logo.svg" alt="Wikidata-logo" /> 
```

resources/js/Components/AuthWidget.vue
guergana: True! Thanks for pointing it out

resources/js/Components/AuthWidget.vue
```diff
@@ -1,11 +1,13 @@
...
+                <img src="../../img/user.svg" class="icon-user" /><span class="username">{{ user.name }}</span>
```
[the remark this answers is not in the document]
itamargiv: Wait, no, I'm a bit of a dunce! Now I see why you were using the source image and the webpack loader. -> for asset versioning. Resolving!

webpack.mix.js
itamargiv: Wait, no, I'm a bit of a dunce! Now I see why you were using the source image and the webpack loader. -> for asset versioning. Resolving!

resources/js/Pages/Layout.vue
itamargiv: Wait, no, I'm a bit of a dunce! Now I see why you were using the source image and the webpack loader. -> for asset versioning. Resolving!

resources/js/Pages/Home.vue
```diff
@@ -49,6 +54,7 @@
                 this.$inertia.get('/results?ids=' + idsToSend)
             },
         },
+        layout: Layout,
```
itamargiv, line 57: I'm not sure this does anything in our case, since we didn't follow https://inertiajs.com/pages#persistent-layouts (we are still using the layout as a component in our template)

public/images/user.svg
guergana: Addressed!

resources/js/Pages/Layout.vue
```diff
@@ -0,0 +1,29 @@
...
+     <h1>{{ $i18n('mismatch-finder-title') }}</h1>
+    <slot />
+  </main>
+</template>
+
+<script lang="ts">
+interface User {
+    username: string,
+    mw_userid: string
+}
```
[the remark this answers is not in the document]
guergana: There is an interface called `User` indeed in two different files: `resources/js/Pages/Layout.vue` and `resources/js/Components/AuthWidget.vue` but they are different. Layout has the values 
```
interface User {
    username: string,
    mw_userid: string
}
```
and AuthWidget has:
```
interface User {
    name: string,
    id: string
}

maybe we should considering naming them differently instead?
```

tests/Vue/Components/AuthWidget.spec.js
```diff
@@ -16,6 +16,6 @@ describe('AuthWidget.vue', () => {
     it('displays a "Guest" when username is missing', () => {
         const wrapper = mount(AuthWidget);
 
-        expect(wrapper.text()).toContain("Guest");
+        expect(wrapper.text()).toContain("Log in");
```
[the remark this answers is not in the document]
guergana: :smile:

resources/js/Pages/Layout.vue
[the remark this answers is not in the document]
itamargiv: I think they should be the same, it's just that the first one is wrong (we made a mistake yesterday while pairing), If you check out the route handler in Laravel, the user object is mapped to the structure of the second interface: https://github.com/wmde/wikidata-mismatch-finder/blob/05980e1f56a52013295971bcaa4b0265fdeef54a/routes/web.php#L22-L24

resources/js/Components/AuthWidget.vue
guergana: I asked @MattiaCapozziWMDE previously for the store and he said it shouldn't but will ask again.

resources/js/Pages/Layout.vue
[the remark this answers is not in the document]
guergana: Thanks for clarifying. Addressed!

resources/js/Components/AuthWidget.vue
guergana: Addressed!

public/i18n/en.json
```diff
@@ -13,5 +13,8 @@
     "item-form-title": "For which Items should the check be performed?",
     "item-form-id-input-label": "Please add one Item identifier per line",
     "item-form-id-input-placeholder": "For example:\nQ345\nQ0987609\nQ4567890",
-    "item-form-submit": "Check Items"
+    "item-form-submit": "Check Items",
+    "log-in": "Log in",
+    "log-out": "Logout",
```
Silvan-WMDE, line 18: :-D

```suggestion
    "log-out": "Log out",
```

Silvan-WMDE (changes requested):

resources/js/Components/AuthWidget.vue
guergana: Leaving as it is for now, after reading @MattiaCapozziWMDE 's input.

resources/js/Pages/Home.vue
guergana: You are right, this is not doiing anything. I've removed it.

itamargiv (approved):
Looking fabulous :nail_care:, Thank you!

itamargiv (changes requested):
hops, wait, sorry seems like it was a premature approval. IT looks great of course still but there are some specs from the ticket that aren't yet met.

itamargiv (approved):
Green! Sorry for the temporary interruption :) Thank you, again!

Silvan-WMDE (approved):

PR #94, opened 2021-09-07
Implement app loading state

itamargiv:
This change introduces [Vuex](https://vuex.vuejs.org/) as a way to manage shared state between page components. This is needed so that we could pass down the global "loading" state from inertia.js into our page components as needed.

As this store setup is adding more code to our app bootstrapping file: `app.ts`. This file is restructured for better readability, and to ensure that  we are only bootstrapping the inertia app if setup succeeds.

Bug: [T290162](https://phabricator.wikimedia.org/T290162)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,8 @@
                 "@wmde/wikit-vue-components": "^2.1.0-alpha.2",
                 "ress": "^3.0.0",
                 "vue": "^2.6.14",
-                "vue-banana-i18n": "^1.5.0"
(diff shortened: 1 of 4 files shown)


Silvan-WMDE (commented):
+1 from me, before @guergana approves it. Looks sophisticated and works great. :+1:

Silvan-WMDE (dismissed):
great work, great conflict resolution - it all seems to work now \o/ :100:

Silvan-WMDE (approved):

PR #95, opened 2021-09-07
Validate TextArea input

Silvan-WMDE:
This change uses the error and warning states of Wikit's TextArea
component to give feedback to the user about the validity of entered
item ids.

Bug: [T290161](https://phabricator.wikimedia.org/T290161)

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,8 +42,8 @@
     "dependencies": {
         "@inertiajs/inertia": "^0.10.1",
         "@inertiajs/inertia-vue": "^0.7.2",
-        "@wmde/wikit-tokens": "^2.1.0-alpha.1",
-        "@wmde/wikit-vue-components": "^2.1.0-alpha.1",
+        "@wmde/wikit-tokens": "^2.1.0-alpha.2",
+        "@wmde/wikit-vue-components": "^2.1.0-alpha.2",
         "ress": "^3.0.0",
         "vue": "^2.6.14",
         "vue-banana-i18n": "^1.5.0"
--- a/resources/js/Pages/Home.vue
+++ b/resources/js/Pages/Home.vue
@@ -13,7 +13,7 @@
                         :label="$i18n('item-form-id-input-label')"
                         :placeholder="$i18n('item-form-id-input-placeholder')"
                         :rows="8"
-                        :error="error ? {message: $i18n(error.message), type: error.type} : null"
+                        :error="error"
                         v-model="form.itemsInput"
                     />
                     <div class="form-buttons">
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -42,8 +42,8 @@
     "dependencies": {
         "@inertiajs/inertia": "^0.10.1",
         "@inertiajs/inertia-vue": "^0.7.2",
-        "@wmde/wikit-tokens": "^2.1.0-alpha.1",
-        "@wmde/wikit-vue-components": "^2.1.0-alpha.1",
+        "@wmde/wikit-tokens": "^2.1.0-alpha.2",
```
itamargiv, line 45: We should have thought of that before ;)

resources/js/Pages/Home.vue
```diff
@@ -12,6 +12,7 @@
                 :label="$i18n('item-form-id-input-label')"
                 :placeholder="$i18n('item-form-id-input-placeholder')"
                 :rows="8"
+                :error="error ? {message: $i18n(error.message), type: error.type} : null"
```
itamargiv, line 15: I think we can simplify this a little, by moving the call to the `$i18n` function into the validation method.

The goal would be to be left with:

```suggestion
                :error="error"
```

resources/js/Pages/Home.vue
```diff
@@ -42,18 +43,51 @@
             user: Object,
         },
         methods: {
+            checkEmpty() {
+                if( !this.form.itemsInput ) {
+                    this.error = {
+                        type: 'warning',
+                        message: 'item-form-error-message-empty'
+                    };
+                    return false;
```
itamargiv, line 52: Nice work! I think, since we are already mutating the internal state of the component, perhaps it is not required to return an explicit boolean? We can just perform the checks and at the end (before sending) check whether the error is set or not.

resources/js/Pages/Home.vue
```diff
@@ -42,18 +43,51 @@
...
+                }
+                return true;
+            },
+            validate() {
+                let valid = true;
+                const lines = this.form.itemsInput.split( '\n' );
+                for ( let line of lines ) {
+                    line = line.trim();
+                    valid = valid && ( line == '' || line.match( /^Q[0-9]*$/ ) );
```
itamargiv, line 61: :heart: Regex! I hate the syntax of it, but it's so versatile for validation.

resources/js/Pages/Home.vue
```diff
@@ -42,18 +43,51 @@
...
+                if( !this.form.itemsInput ) {
+                    this.error = {
+                        type: 'warning',
+                        message: 'item-form-error-message-empty'
+                    };
+                    return false;
+                }
+                return true;
+            },
+            validate() {
```
itamargiv, line 56: So I think this method could probably be split into two or three: one to deserialize the IDs into an array, one to perform checks and set the error, and another one to serialize the IDs into the pipe separated format.

That way not only are you ensuring each function has its own responsibility to take care of, but you are also decoupling the validation and sanitization logic from one another.

resources/js/Pages/Home.vue
```diff
@@ -42,18 +43,51 @@
...
+                let valid = true;
+                const lines = this.form.itemsInput.split( '\n' );
+                for ( let line of lines ) {
```
itamargiv, line 59: There are some spiffy JS higher order functions for checks like this, namely: [`Array.prototype.every()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every). Of course, this is optional, and the implementers choice :smile:

resources/js/Pages/Home.vue
```diff
@@ -42,18 +43,51 @@
             user: Object,
         },
         methods: {
+            checkEmpty() {
+                if( !this.form.itemsInput ) {
+                    this.error = {
+                        type: 'warning',
+                        message: 'item-form-error-message-empty'
```
itamargiv, line 50: To continue the first comment made, I think you might be able to use the i18n function here like so (though I didn't try it out):

```suggestion
                        message: this.$i18n('item-form-error-message-empty')
```

resources/js/Pages/Home.vue
```diff
@@ -42,18 +43,51 @@
...
+                if( !this.checkEmpty() ) {
+                    return;
+                }
 
-                const idsToSend = this.form.itemsInput.split('\n').join('|');
-                this.$inertia.get('/results?ids=' + idsToSend)
+                const validatedIds = this.validate();
+                if( validatedIds ) {
+                    this.$inertia.get( '/results?ids=' + validatedIds );
+                }
```
itamargiv, line 82: > Since we are already mutating the internal state of the component...

To build up on that idea, if we were just to check the error prop in this function it would reduce the b`if branching here to basically one, something like this:

```js
send() {
    this.validate();

    if (this.error) {
        return;
    }
   
    this.$inertia.get( '/results?ids=' + validatedIds );
}

**Review by itamargiv — CHANGES_REQUESTED** [review]
Fantastic work, I left a couple of comments, but the logic is sound, it's really just some ideas on how to reduce lines and branches.

Also, I think it will be worthwhile to add some browser tests to cover the BDD flows in the parent (uncle, maybe) ticket [T287951](https://phabricator.wikimedia.org/T287951):

File: resources/js/Pages/Home.vue (hunk 8 above)
Comment (Silvan-WMDE) [reply resources/js/Pages/Home.vue]: Nice, I like this. I also tried the computed properties, but it looks like they are not really needed here.

File: tests/Browser/ItemsFormTest.php [hunk 9]
```diff
@@ -34,4 +34,39 @@ public function test_can_submit_list_of_item_ids()
...
+    {
+        $this->browse(function (Browser $browser) {
+            $browser->visit(new HomePage)
+                    ->press('button')
+                    ->assertSee('Please provide the Item identifiers in order to perform the check.')
+                    ->keys('@items-input', 'Q1', '{return_key}', 'Q2')
+                    ->press('button')
+                    ->assertDontSee('Please provide the Item identifiers in order to perform the check.');
+        });
+    }
```
itamargiv, line 59: Nice addition!

itamargiv (approved):
Thank you for the changes, great work! :+1:

PR #98, opened 2021-09-08
Ensure type checks in all page components

itamargiv:
This change ensures that all page components are type checked, and corrects a few previously undiscovered type checking issues.


guergana (approved):
This looks like approved to me! :heavy_check_mark:

PR #100, opened 2021-09-10
Fix sizing of website logo in mobile version

guergana:
Since some of the header styles are shared with the header in the store, check as well there when performing the review.

### DIFF
--- a/resources/sass/app.scss
+++ b/resources/sass/app.scss
@@ -24,12 +24,7 @@ section {
 }
 
 h1, .h1 {
-    margin: $dimension-layout-xsmall 0;
     @include heading-1;
-
-    @media(min-width: $width-breakpoint-tablet){
-        margin: $dimension-layout-small 0;
-    }
 }
 
 h4, .h4 {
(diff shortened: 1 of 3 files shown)


Silvan-WMDE (approved):
Looks good to me - thank you!

resources/sass/app.scss
```diff
@@ -61,19 +56,46 @@ a {
...
+        @media (max-width: $width-breakpoint-tablet) {
+            flex-direction: column;
+        }
+    }
+    h1 {
+        @media (max-width: $width-breakpoint-tablet) {
+            margin: $dimension-layout-xxsmall 0;
+        }
+    }
+}
```
itamargiv, line 89: Instead of putting this in the global sass file, I think it would be best if we put it where it is actually used, In the Layout.vue file

resources/sass/app.scss
```diff
@@ -24,12 +24,7 @@ section {
 }
 
 h1, .h1 {
-    margin: $dimension-layout-xsmall 0;
     @include heading-1;
-
-    @media(min-width: $width-breakpoint-tablet){
-        margin: $dimension-layout-small 0;
-    }
```
itamargiv, line 32: Why are we moving these rules from here? Should the store and the website not have the same spacing in the heading?

PR #103, opened 2021-09-14
Fix header tokens

itamargiv:
This change reverts the headers token back to their correct assignment as based on the Mismatch Finder website's [figma specification](https://www.figma.com/file/sgK1Go5my7ShZ1HQIxV2FI/Mismatch-Finder?node-id=1119%3A22624) and to ensure the consistency of our global styles.

In addition, the WiKit version is updated in order to ensure the correct sizing of layout tokens, as they were corrected in WiKit itself.

Bug: [T290634](https://phabricator.wikimedia.org/T290634)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2950,17 +2950,17 @@
             }
         },
         "node_modules/@wmde/wikit-tokens": {
-            "version": "2.1.0-alpha.2",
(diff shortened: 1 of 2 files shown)


guergana (approved):
LGTM

PR #104, opened 2021-09-14
Apply QID validation on results web route

itamargiv:
This change applies to the `results` web route, the same validation and sanitization logic that we apply to our `mismatches` api route.


Silvan-WMDE (approved):
nice and easier than I thought :+1:

PR #105, opened 2021-09-16
Add stateful browser tests and fix CI to run them

Silvan-WMDE:
This change adds a few browser tests from #70 and configures GitHub actions in such a way, that `dusk` has a working database connection and can run stateful browser test.

Bug: [T289154](https://phabricator.wikimedia.org/T289154)


itamargiv (approved):
Looking good! Thank you!

PR #106, opened 2021-09-16
Pass json response from backend into the Results Page component

guergana:
This PR shows the requested results by id from the web endpoint `/results?ids=Q1|Q2` in the Results page.

### DIFF
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -19,7 +19,7 @@
                 </tr>
             </tbody>
         </table>
-        <p v-else class="results">
+        <p v-else class="not-found">
             Thank you for sending IDs {{item_ids}}. 
             The requested item ids didn't match any entries in our database. 
             Please try with a different set of ids.
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -5,6 +5,23 @@
         <p class="results">
             Thank you for sending IDs {{item_ids}}. Mismatch results will be displayed in this page soon.
         </p>
+        <table>
+            <tbody>
+                <tr v-for="result in results" :key="result.id">
+                    <td>{{result.id}}</td>
+                    <td>{{result.item_id}}</td>
+                    <td>{{result.statement_guid}}</td>
+                    <td>{{result.property_id}}</td>
+                    <td>{{result.wikidata_value}}</td>
+                    <td>{{result.external_value}}</td>
+                    <td>{{result.external_url}}</td>
+                    <td>{{result.review_status}}</td>
+                    <td>{{result.reviewer}}</td>
+                    <td>{{result.import}}</td>
+                    <td>{{result.updated_at}}</td>
+                </tr>
+            </tbody>
+        </table>
     </div>
 </template>
 
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -35,7 +35,8 @@
             Head
         },
         props: {
-            item_ids: Array
+            item_ids: Array,
+            results: Array
         }
     });
 </script>
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -2,9 +2,6 @@
     <div>
         <Head title="Mismatch Finder - Results" />
         <h2>Results</h2>
-        <p class="results">
-            Thank you for sending IDs {{item_ids}}. Mismatch results will be displayed in this page soon.
-        </p>
         <table>
             <tbody>
                 <tr v-for="result in results" :key="result.id">
@@ -35,7 +32,6 @@
             Head
         },
         props: {
-            item_ids: Array,
             results: Array
         }
     });
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -2,7 +2,7 @@
     <div>
         <Head title="Mismatch Finder - Results" />
         <h2>Results</h2>
-        <table>
+        <table v-if="results.length">
             <tbody>
                 <tr v-for="result in results" :key="result.id">
                     <td>{{result.id}}</td>
@@ -19,6 +19,11 @@
                 </tr>
             </tbody>
         </table>
+        <p v-else class="results">
+            Thank you for sending IDs {{item_ids}}. 
+            The requested item ids didn't match any entries in our database. 
+            Please try with a different set of ids.
+        </p>
     </div>
 </template>
 
@@ -32,6 +37,7 @@
             Head
         },
         props: {
+            item_ids: Array,
             results: Array
         }
     });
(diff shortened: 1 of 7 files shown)


routes/web.php
```diff
@@ -36,9 +35,12 @@
 
     $ids = $request->input('ids');
 
+    $results = app('app\Http\Controllers\MismatchController')->index($request);
```
itamargiv, line 38: :star_struck: NICE!! :star_struck:

Silvan-WMDE (dismissed):
Yay, this does what it's supposed to do and browser tests now pass again. \o/

I still left a few comments for discussion below. Also not much is being tested at the moment. I suggested an inertia test in #109 to make sure we actually pass the required data into the Results page.

resources/js/Pages/Results.vue
```diff
@@ -5,6 +5,23 @@
         <p class="results">
             Thank you for sending IDs {{item_ids}}. Mismatch results will be displayed in this page soon.
         </p>
+        <table>
+            <tbody>
+                <tr v-for="result in results" :key="result.id">
```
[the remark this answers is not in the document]
guergana: We don't. But i thought it was a good idea to have visual proof..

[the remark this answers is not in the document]
guergana: I don't think a test is really needed, this can be introduced later with the table component.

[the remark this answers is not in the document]
Silvan-WMDE: right, fine :+1:

routes/web.php
guergana: I am not sure about what could be good alternatives. Maybe @itamargiv or @mariushoch  could offer a suggestion?

itamargiv: To clarify, I love this approach as it seems so easy, and I think it's a nice original, and pretty elegant solution that reuses already existing code. I must say, though, that I left this comment without doing a full review, so it is a bit out of context.

@Silvan-WMDE is of course right, the logic for the web endpoint and the logic for the API endpoint are not fully compatible. And we perhaps don't necessarily want to rely on the results of the API endpoint for this specific view of the data in the database. Therefore, the only solution I can think of is that you query the database in this route handler as well, especially since we still have some other operations to perform on the data before passing it to the page, such as fetching entity labels from the Wikidata API.

guergana: Very good points! I've updated the PR following @Silvan-WMDE 's suggestion.

routes/web.php
```diff
@@ -36,9 +35,12 @@
...
+    
     return inertia('Results', [
         'item_ids' => $ids,
```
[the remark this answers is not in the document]
guergana: Removed.

Silvan-WMDE (commented):
Still works. :+1:
I left a few comments on the new browser tests below, but anyway they will have to be replaced by tests for the table component soon, I guess.

Also, I still belive that the inertia test from #109 would make a lot of sense here, as it actually checks that the mismatch data is passed into the component.

tests/Browser/ItemsFormTest.php
```diff
@@ -23,15 +29,62 @@ public function test_can_enter_list_of_item_ids()
...
+
+            Mismatch::factory(2)
+                ->for($import)
+                ->create();
+            
+            $mismatch = Mismatch::first();
+            $mismatch2 = Mismatch::find(2);
+
+            $item1_in_db_id =  $mismatch->item_id;
+            $item2_in_db_id =  $mismatch2->item_id;
```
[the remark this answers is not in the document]
guergana: feel free to change it ;)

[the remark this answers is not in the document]
itamargiv: I would also be with @Silvan-WMDE on this. But this means that we need some kind of linter rule for it, and since we don't have one, I think it's fine for now.

tests/Browser/ItemsFormTest.php
```diff
@@ -23,15 +29,46 @@ public function test_can_enter_list_of_item_ids()
...
+            $import = ImportMeta::factory()
+            ->for(User::factory()->uploader())
+            ->create();
+
+            Mismatch::factory(2)
+                ->for($import)
+                ->create();
+            
+            $mismatch = Mismatch::first();
+            $mismatch2 = Mismatch::find(2);
```
itamargiv, line 60: I think you can just use `Mismatch::all()` to retrieve all mismatches.

tests/Browser/ItemsFormTest.php
```diff
@@ -23,15 +29,46 @@ public function test_can_enter_list_of_item_ids()
...
             $browser->visit(new HomePage)
-                    ->keys('@items-input', 'Q1', '{return_key}', 'q2')
+                    ->keys('@items-input', $item1_in_db_id, '{return_key}', $item2_in_db_id)
                     ->press('button')
-                    ->waitFor('@results')
+                    ->waitFor('table')
                     ->assertTitle('Mismatch Finder - Results')
-                    ->assertSee('[ "Q1", "Q2" ]');
+                    ->assertSee($item1_in_db_id)
+                    ->assertSee($item2_in_db_id);
```
itamargiv, line 71: Nice test! :100:

tests/Feature/WebRouteTest.php
```diff
@@ -81,6 +81,36 @@ public function test_results_route()
...
+
+        $response = $this->get(route('results', [
+            'ids' => $mismatch->item_id
+        ]));
+
+        $response->assertSuccessful();
+        $response->assertViewIs('app')
+            ->assertInertia(function (Assert $page) use ($mismatch, $import) {
+                $page->component('Results')
+                    ->has('results.0', 11) // result object size
```
itamargiv, line 107: Non-blocking, but I'm not sure if this is needed. We care that it shows particular parts of the data, if we tie this to the size of a mismatch object it will make the test more brittle.

tests/Browser/ItemsFormTest.php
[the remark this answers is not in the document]
guergana: i subconsciously emulated the naming convention of the test functions. I don't really know what the convention for multi-variable names in laravel is.

itamargiv (approved):
Nice work, I left a few comments, but generally speaking I think this can be merged, I will be changing these tests that I commented on anyway while working on my current task.

tests/Feature/WebRouteTest.php
guergana: Yes, this could change. I was also not sure about using the number.

guergana:
This PR shows the requested results by id from the web endpoint /results?ids=Q1|Q2 in the Results page and shows them in a rudimentary table element that will be replaced in a subsequent PR.

Bug: [T291064](https://phabricator.wikimedia.org/T291064)

tests/Feature/WebRouteTest.php
Silvan-WMDE: see the discussion we had on the size test in the other PR: https://github.com/wmde/wikidata-mismatch-finder/pull/109#pullrequestreview-758480808

PR #107, opened 2021-09-16
change example Item ID in error message

lydiapintscher:
more meaningful and more in line with the placeholder IDs


itamargiv (approved):
:pray: Thank you!

PR #108, opened 2021-09-16
Add basic table component

itamargiv:
This change adds the basic table component required to display mismatch results for our table. This component serves as a wrapper for the generic HTML5 table components, in order to apply styles and responsive fallback to displayed tabular data. See `Playground.vue` for a usage example.

Bug: [T289843](https://phabricator.wikimedia.org/T289843)

### DIFF
--- a/resources/js/Components/Table.vue
+++ b/resources/js/Components/Table.vue
@@ -25,6 +25,62 @@ export default Vue.extend({
 <style lang="scss">
     @import '~@wmde/wikit-tokens/dist/_variables.scss';
 
+    @mixin LinearTable {
+        /**
+        * Wipe the thead from the face of the earth
+        * modern screen readers will expose the
+        * generated content
+        */
+        thead {
+            display: none;
+            visibility: hidden;
+        }
+
+        /**
+        * make everything display flex for alignment
+        */
+        tbody, tr, th, td {
+            height: auto;
+            display: flex;
+            flex-direction: column;
+        }
+
+        td, th {
+            flex-direction: row;
+            flex-basis: 60%;
+        }
+
+        /**
+        * Labeling
+        *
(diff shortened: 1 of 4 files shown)


resources/js/Components/Table.vue
```diff
@@ -0,0 +1,180 @@
...
+
+        th {
+            /**
+            * Typography
+            */
+            font-weight: $font-weight-bold;
+        }
+
+        &--linear-mobile {
+            @media (max-width: $width-breakpoint-mobile){
```
guergana, line 163: Awesome!!! It looks great but I would let someone from UX take a look. :)

resources/js/Pages/Playground.vue
```diff
@@ -1,27 +1,37 @@
...
+            <tr>
+                <th scope="col">Name</th>
+                <th scope="col">Calories</th>
+            </tr>
+        </thead>
+        <tbody>
+            <tr v-for="dessert in desserts" :key="dessert.name">
+                <!-- When using the table make sure to include the
+                    data-header for linearization and accesibility -->
+                <td data-header="Name">{{dessert.name}}</td>
```
guergana, line 18: it is not immediately obvious to me if we can pass anything other than text into this component. wouldn't it make sense to have an example in the playground that introduces some other element in a cell? we will need this right away in the implementation for MF. What do you think?

Silvan-WMDE: I added an example to the Playground, that illustrates how to use the Link component with a url from the data section.

Silvan-WMDE (approved):
Looks great and works like a charm. :100:

resources/js/Pages/Playground.vue
guergana: Great. Thanks for adding the example.

PR #109, opened 2021-09-17
Suggested test

Silvan-WMDE:


tests/Feature/WebRouteTest.php
```diff
@@ -81,6 +83,36 @@ public function test_results_route()
...
+
+        $response = $this->get(route('results', [
+            'ids' => $mismatch->item_id
+        ]));
+
+        $response->assertSuccessful();
+        $response->assertViewIs('app')
+            ->assertInertia(function (Assert $page) use ($mismatch, $import) {
+                $page->component('Results')
+                    ->has('results.0', 11)
```
guergana, line 109: Hello, where is that 11 coming from???

Silvan-WMDE: that's just the size of the results object, I counted the number of properties

Silvan-WMDE: Sure, feel free - it's just a suggestion. From my experience it makes sense to pinpoint the object size in such a test, because once something changes, the test will fail and indicate that maybe other modifications to tests are needed, too. Also the error message is straightforward: `Inertia property [results.0] does not have the expected size.`

guergana: Could you add the comment? Then I can approve. Thanks :)

guergana: nevermind. rebase is not needed

PR #110, opened 2021-09-22
Client/home page error handling

Silvan-WMDE:
This change redirects any error response by the server back to the home
route and adds an 'errors' flash message to the Inertia response. This
way the Home page can pick up the flash message and indicate a server
error with a Message component, triggered by a computed prop.

Testers will be able to fake/force a server error by sending a
'X-Mismatch-Results-Error' header field with their request to the
results route.

Bug: [T289344](https://phabricator.wikimedia.org/T289344)

### DIFF
--- a/app/Http/Middleware/SimulateError.php
+++ b/app/Http/Middleware/SimulateError.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace App\Http\Middleware;
+
+use Closure;
+use Illuminate\Http\Request;
+use RuntimeException;
+
+/**
+ * This middleware will force a RuntimeException for testing purposes, if the
+ * 'X-Mismatch-Results-Error' HTTP header field is sent with the request. It
+ * allows for simple feature tests of the error routing behaviour as well as
+ * easy verification by testers, using a browser extension such as
+ * https://addons.mozilla.org/de/firefox/addon/modheader-firefox/
+ */
+class SimulateError
+{
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle(Request $request, Closure $next)
+    {
+        if ($request->header('X-Mismatch-Results-Error')) {
+            throw new RuntimeException("Simulated Server Error");
+        }
+    
+        return $next($request);
+    }
+}
--- a/resources/js/Pages/Home.vue
+++ b/resources/js/Pages/Home.vue
@@ -6,9 +6,11 @@
             <p id="about-description" >{{ $i18n('about-mismatch-finder-description') }}</p>
         </section>
 
-        <Message id="error-message" v-if="unexpectedError" type="error">
-            <span>{{ $i18n('mismatch-query-server-error') }}</span>
-        </Message>
+        <section id="message-section">
+            <Message v-if="unexpectedError" type="error">
+                <span>{{ $i18n('mismatch-query-server-error') }}</span>
+            </Message>
+        </section>
 
         <section id="querying-section">
             <h2 class="h5">{{ $i18n('item-form-title') }}</h2>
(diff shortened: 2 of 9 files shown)


routes/web.php
```diff
@@ -30,6 +30,11 @@
 })->name('home');
 
 Route::get('/results', function (MismatchGetRequest $request) {
+    if ($request->header('X-Mismatch-Results-Error')) {
+        // force a 500 server response for testing purposes
+        throw new RuntimeException("Simulated Server Error");
+    }
```
itamargiv, line 36: I'm not sure if we should leave this in the code. But if we do, I'd rather we move this out of the route handler, and maybe create some kind of global middleware for it. In the same way we did for the uselang parameter: https://github.com/wmde/wikidata-mismatch-finder/blob/main/app/Http/Middleware/SetLocale.php

tests/Feature/WebRouteTest.php
```diff
@@ -162,4 +162,25 @@ public function test_store_import_status()
...
+                ['X-Mismatch-Results-Error' => 'true']
+            );
+
+        $redirect->assertRedirect(route('home'));
+        $response = $this->get($redirect->headers->get('Location'));
+        $response->assertInertia(function (Assert $page) {
+                $page->component('Home')
+                    ->where('flash.errors', [ 'unexpected' => 'Unexpected error']);
+        });
+    }
```
itamargiv, line 185: You were right, I love this! I'm really impressed with all the testing helpers Laravel and inertia provide us with :100:

itamargiv (changes requested):
This looks fantastic, Thank you! :+1: I left a couple of comment regarding style and document structure. Also, I'm not sure about this Header intervention mechanism, it's neat, but we probably only need it for dev environments?

resources/js/Pages/Home.vue
```diff
@@ -5,6 +5,11 @@
             <h2 class="h4">{{ $i18n('about-mismatch-finder-title') }}</h2>
             <p id="about-description" >{{ $i18n('about-mismatch-finder-description') }}</p>
         </section>
+
+        <Message id="error-message" v-if="unexpectedError" type="error">
+            <span>{{ $i18n('mismatch-query-server-error') }}</span>
+        </Message>
+
```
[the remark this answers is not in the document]
Silvan-WMDE: Yes, a section of its own it is now.

resources/js/Pages/Home.vue
```diff
@@ -150,4 +168,10 @@
         text-align: end;
     }
 }
+
+#error-message {
+    max-width: 675px;
+    margin: 40px 0px 40px 0px;
+    border-radius: $border-radius-base;
+}
```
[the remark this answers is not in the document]
Silvan-WMDE: The `margin` is solved by putting it into its own section, `max-width` remains hard-coded until we have discussed a possible token solution with @MattiaCapozziWMDE.

routes/web.php
Silvan-WMDE: moved

tests/Feature/WebRouteTest.php
Silvan-WMDE: added a comment to make more explicit that we first get a redirect response, which we then follow to the URL from the 'Location' header.

app/Http/Middleware/SimulateError.php
```diff
@@ -0,0 +1,33 @@
...
+     */
+    public function handle(Request $request, Closure $next)
+    {
+        if ($request->header('X-Mismatch-Results-Error')) {
+            throw new RuntimeException("Simulated Server Error");
+        }
+    
+        return $next($request);
+    }
+}
```
itamargiv, line 33: Sometimes you just have to appreciate how pretty OOP patterns can be, verbosity and all :bouquet:

routes/web.php
```diff
@@ -30,22 +30,23 @@
     ]);
 })->name('home');
 
-Route::get('/results', function (MismatchGetRequest $request) {
-    $user = Auth::user() ? [
-        'name' => Auth::user()->username
-    ] : null;
+Route::middleware('simulateError')
```
itamargiv, line 33: Nicely done!

itamargiv (approved):
Well done! Great job, and welcome to the Frontend! :raised_hands:

PR #112, opened 2021-09-24
Update route handler to group results by QID

itamargiv:
This change updates the mismatch results route handler to provide mismatches grouped by item ID. This is required for displaying mismatches per item, and determining which items did not provide any mismatch results.

Bug: [T291569](https://phabricator.wikimedia.org/T291569)
Bug: [T289559](https://phabricator.wikimedia.org/T289559)

### DIFF
--- a/tests/Feature/WebRouteTest.php
+++ b/tests/Feature/WebRouteTest.php
@@ -96,19 +96,31 @@ class WebRouteTest extends TestCase
             ->for($import)
             ->create();
 
+        $qid = $mismatch->item_id;
+
+        $isMismatch = function (Assert $data) use ($mismatch, $import) {
+            $data->whereAll([
+                'id' => $mismatch->id,
+                // Casting values to string, as it seems that the inertia
+                // testing helper also converts all values to strings
+                'property_id' => (string) $mismatch->property_id,
+                'wikidata_value' => (string) $mismatch->wikidata_value,
+                'external_value' => (string) $mismatch->external_value,
+                'import_meta.user.username' => (string) $import->user->username,
+                'import_meta.created_at' => $import->created_at->toISOString()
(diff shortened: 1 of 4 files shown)


tests/Feature/WebRouteTest.php
```diff
@@ -96,19 +96,31 @@ public function test_results_route_retrieves_mismatches()
...
+                'property_id' => (string) $mismatch->property_id,
+                'wikidata_value' => (string) $mismatch->wikidata_value,
+                'external_value' => (string) $mismatch->external_value,
+                'import_meta.user.username' => (string) $import->user->username,
+                'import_meta.created_at' => $import->created_at->toISOString()
+            ])->etc();
+        };
+
+        $withResultsPage = function (Assert $page) use ($qid, $isMismatch) {
+            $page->component('Results')->has("results.$qid.0", $isMismatch);
```
Silvan-WMDE, line 115: So `results.$qid.0` is the representation of the grouping, right? I actually would have expected a test that sends two ids and then asserts that they are both there, grouped, to show it more clearly. But maybe you are right and grouping doesn't actually need multiple results to work.

itamargiv: Essentially, the way Laravel groups collections, it is not necessarily required, as it adds an additional level to the array into  is keyed by the value of the group. Therefore:

**Before**
```php
[
  [
    'item_id' => 'Q1'
    //...
  ],
  //...
]
```

**After**
```php
[
  'Q1' => [
      [
        'item_id' => 'Q1'
        //...
      ],
      //...
  ],
  //...
]
```

Silvan-WMDE (approved):
Looks good, works locally, is needed for next steps, so let's get this merged...

PR #114, opened 2021-09-24
Add labels and links

itamargiv:
_**Note:** This change depends on #113_

This change adds an entity ID to Wikidata label map to the results page. This map is then used to display labels for various entity IDs displayed on the page.

Bug: [T291569](https://phabricator.wikimedia.org/T291569)

### DIFF
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -3,7 +3,9 @@
         <Head title="Mismatch Finder - Results" />
         <section id="results" v-if="Object.keys(results).length">
             <section class="item-mismatches" v-for="(mismatches, item, idx) in results" :key="idx">
-                <h2 class="h4">{{item}}</h2>
+                <h2 class="h4">
+                    <wikit-link :href="`http://www.wikidata.org/wiki/${item}`">{{labels[item]}} ({{item}})</wikit-link>
+                </h2>
                 <mismatches-table :mismatches="mapLabels(mismatches)" />
             </section>
         </section>
@@ -19,26 +21,47 @@
     import { PropType } from 'vue';
 
     import { Head } from '@inertiajs/inertia-vue';
+    import { Link as WikitLink } from '@wmde/wikit-vue-components';
 
     import MismatchesTable from '../Components/MismatchesTable.vue';
-    import Mismatch from '../types/Mismatch';
+    import Mismatch, {LabelledMismatch} from '../types/Mismatch';
     import defineComponent from '../types/defineComponent';
 
     interface Result {
         [qid: string]: Mismatch[]
     }
 
+    interface LabelMap {
+        [entityId: string]: string
+    }
+
     export default defineComponent({
         components: {
             Head,
-            MismatchesTable
+            MismatchesTable,
+            WikitLink
         },
         props: {
             item_ids: Array as PropType<string[]>,
-            results: Object as PropType<Result>
+            results: Object as PropType<Result>,
+            labels: Object as PropType<LabelMap>
+        },
+        methods: {
+            mapLabels(mismatches: Mismatch[]): LabelledMismatch[]{
+                return mismatches.map(mismatch => ({
+                    property_label: this.labels[mismatch.property_id],
+                    value_label: this.labels[mismatch.wikidata_value] || null,
+                    ...mismatch
+                }));
+            }
         }
     });
 </script>
 
 <style lang="scss">
+    h2 {
+        .wikit-Link.wikit {
+            font-weight: bold;
+        }
+    }
 </style>
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -6,7 +6,7 @@
                 <h2 class="h4">
                     <wikit-link :href="`http://www.wikidata.org/wiki/${item}`">{{labels[item]}} ({{item}})</wikit-link>
                 </h2>
-                <mismatches-table :mismatches="mapLabels(mismatches)" />
+                <mismatches-table :mismatches="addLabels(mismatches)" />
             </section>
         </section>
         <p v-else class="not-found">
@@ -47,7 +47,10 @@
             labels: Object as PropType<LabelMap>
         },
         methods: {
-            mapLabels(mismatches: Mismatch[]): LabelledMismatch[]{
+            addLabels(mismatches: Mismatch[]): LabelledMismatch[]{
+                // The following callback maps existing mismatches to extended
+                // mismatch objects which include labels, by looking up any
+                // potential entity ids within the labels object.
                 return mismatches.map(mismatch => ({
                     property_label: this.labels[mismatch.property_id],
                     value_label: this.labels[mismatch.wikidata_value] || null,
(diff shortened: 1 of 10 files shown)


routes/web.php
```diff
@@ -30,22 +32,37 @@
...
+            if (!in_array($mismatch->property_id, $ids)) {
+                $ids[] = $mismatch->property_id;
+            }
+
+            if ($entityValue && !in_array($wikidataValue, $ids)) {
+                $ids[] = $wikidataValue;
+            }
+
+            return $ids;
+        }, $itemIds);
```
Silvan-WMDE, line 59: Ok, just to summarise:
this `reduce` function iterates through all the mismatches and collects entityIds for which labels need to be fetched. It also starts with the `$itemIds` as initial value, which means that labels will be fetched for all Q-ids from the input field, regardless of whether mismatches for them can be found in the db or not. Is that right?

I think at least this last point deserves a comment in the code, if not a quick overall explanation as to what the combination of `$entities`, `$mismatches`, the `reduce` function and the `$itemIds` argument will actually achieve.

Also, this is getting quite complex and I'm wondering at which point it actually makes sense to move it all into its own `ResultsController`, rather than adding more and more functionality to a single ->get()` method in the web.php file. Collecting entity IDs and sending them to Wikidata to fetch labels may actually deserve its own method (or two), doesn't it?

tests/Feature/WebRouteTest.php
```diff
@@ -103,6 +103,8 @@ public function test_results_route_retrieves_mismatches()
                 'id' => $mismatch->id,
                 // Casting values to string, as it seems that the inertia
                 // testing helper also converts all values to strings
+                'item_id' => (string) $mismatch->item_id,
+                'statement_guid' => (string) $mismatch->statement_guid,
```
Silvan-WMDE, line 107: these two are not actually new, they've always been there, right? It's just that now they are needed, and therefore you newly include them in the test?

routes/web.php
```diff
@@ -30,22 +32,37 @@
...
 
         return inertia('Results', [
-            'item_ids' => $ids,
             'user' => $user,
-            'results' => $results,
+            'item_ids' => $itemIds,
+            'results' => $mismatches->groupBy('item_id'),
+            'labels' => $wikidata->getLabels($entityIds, App::getLocale())
```
Silvan-WMDE, line 65: this whole new `labels` array is added here but never tested for in ẀebRouteTest.php` below, I think.

resources/js/Pages/Results.vue
```diff
@@ -19,26 +21,47 @@
...
+            results: Object as PropType<Result>,
+            labels: Object as PropType<LabelMap>
+        },
+        methods: {
+            mapLabels(mismatches: Mismatch[]): LabelledMismatch[]{
+                return mismatches.map(mismatch => ({
+                    property_label: this.labels[mismatch.property_id],
+                    value_label: this.labels[mismatch.wikidata_value] || null,
+                    ...mismatch
+                }));
```
Silvan-WMDE, line 55: ok, this looks equally scary as the `reduce` function below, but basically adds a `property_label` and `value_label` field (or `null`, if not applicable?) to each mismatch, while leaving the rest of the mismatch fields untouched, right? Nice. Maybe also worth a comment? Or do you think the code speaks for itself here?

Also, the method name: wouldn't `addLabels` actually be a better fit here, because `map` is the really technical activity of iterating through the mismatches and applying some function to each element, rather than the conventional "map A to B"?

Silvan-WMDE (changes requested):
Hey, this is fanatastic work. It adds so much functionality into short and precise code, I really like it. just added some comments below, almost only for my own understanding.

Still, I think I found some tests missing, and I guess if we had browser tests for the results page, then they would have to be modified a little with this change? Shouldn't we add some?

tests/Feature/WebRouteTest.php
guergana: I also noticed this

routes/web.php
itamargiv: > [...], which means that labels will be fetched for all Q-ids from the input field, regardless of whether mismatches for them can be found in the db or not. Is that right?

Correct!

> I think at least this last point deserves a comment in the code, if not a quick overall explanation as to what the combination of $entities, $mismatches, the reduce function and the $itemIds argument will actually achieve.

Good point, will do.

> Collecting entity IDs and sending them to Wikidata to fetch labels may actually deserve its own method (or two), doesn't it?

Yes, agreed.

tests/Feature/WebRouteTest.php
itamargiv: Yes indeed, as long as I don't need them, there is no need to test for them, otherwise tests will become brittle.

routes/web.php
itamargiv: Good point, I will update the tests.

resources/js/Pages/Results.vue
itamargiv: > Maybe also worth a comment? Or do you think the code speaks for itself here?

To me the code speaks for itself, this is a pretty common pattern in JS. But, I'm not a big fan of implicit knowledge, so if you feel like a comment is needed, that is reason enough for me to add one :smiley: 

> Also, the method name: wouldn't addLabels actually be a better fit here

Sure, yeah, why not all for clarity.

routes/web.php
itamargiv: Done

routes/web.php
itamargiv: Done

resources/js/Pages/Results.vue
itamargiv: Done

itamargiv:
Hey hey, thank you for the thorough review I updated the code according to your remarks, added more comments so that hopefully the code is clearer now. I also updated the browser tests, and found a bug with Qids that start with a lowercase Q (see last commit). Feel free to unresolve any of the comments if there is more to add or change.

routes/web.php
```diff
@@ -30,24 +33,8 @@
...
-            ->groupBy('item_id');
-
-        return inertia('Results', [
-            'item_ids' => $ids,
-            'user' => $user,
-            'results' => $results,
-        ]);
-    })->name('results');
+    ->get('/results', [ResultsController::class, 'index'])
+    ->name('results');
```
Silvan-WMDE, line 37: ah, doesn't it look great? :-)

tests/Vue/Components/MismatchRow.spec.js
```diff
@@ -27,11 +26,35 @@ describe('MismatchesRow.vue', () => {
...
-            mismatch.property_id,
+            mismatch.property_label,
             mismatch.wikidata_value,
             mismatch.external_value,
             mismatch.import_meta.user.username,
             mismatch.import_meta.created_at
         );
     });
+
+    it('shows wikidata label over value when available', () => {
```
Silvan-WMDE, line 37: nice addition

Silvan-WMDE (approved):
Thank you for all the changes! :100:

Silvan-WMDE (approved):

tests/Vue/Components/MismatchRow.spec.js
itamargiv: Thanks :)

routes/web.php
itamargiv: It truely does

itamargiv:
Thanks for the typo fix!

PR #115, opened 2021-09-24
Update results page with no mismatches available for review

guergana:
This PR displays a message indicating when there are no mismatches in the store for the Item IDs the user provides. 

Bug: [T289559](https://phabricator.wikimedia.org/T289559)

### DIFF
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -25,7 +25,6 @@
     "column-external-value": "Value on external source",
     "column-upload-info": "Upload Info",
     "no-mismatches-found-message": "No mismatches have been found for:",
-    "refine-items-selection": "Refine Items Selection",
     "results-page-description": "You can fix them! The table below presents you with links to the statement on Wikidata and to the entry on the external source. You can compare the values and manually fix the mismatch, editing Wikidata or the external data source. After investigating and fixing the mismatch, you can indicate their status using the options from the drop-down.",
     "results-page-title": "What should I do with the mismatches?"
 }
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -26,6 +26,6 @@
     "column-upload-info": "Upload Info",
     "no-mismatches-found-message": "No mismatches have been found for:",
     "refine-items-selection": "Refine Items Selection",
-    "results-page-description": "You can fix them! The table below presents you with links to the statement on Wikidata and to the entry on the external source. You can compare the values and manually fix the mismatch, editing Wikidata or the external data source. After investigating and fixing the mismatch, you can indicate their status using the options from the drop-down. Consult the instructions",
+    "results-page-description": "You can fix them! The table below presents you with links to the statement on Wikidata and to the entry on the external source. You can compare the values and manually fix the mismatch, editing Wikidata or the external data source. After investigating and fixing the mismatch, you can indicate their status using the options from the drop-down.",
     "results-page-title": "What should I do with the mismatches?"
 }
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -24,6 +24,5 @@
     "column-wikidata-value": "Value on Wikidata",
     "column-external-value": "Value on external source",
     "column-upload-info": "Upload Info",
-    "no-mismatches-found-message": "No mismatches have been found for:",
-    "results-page-title": "What should I do with the mismatches?"
+    "no-mismatches-found-message": "No mismatches have been found for:"
 }
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -23,5 +23,9 @@
     "column-property": "Property",
     "column-wikidata-value": "Value on Wikidata",
     "column-external-value": "Value on external source",
-    "column-upload-info": "Upload Info"
+    "column-upload-info": "Upload Info",
+    "no-mismatches-found-message": "No mismatches have been found for:",
+    "refine-items-selection": "Refine Items Selection",
+    "results-page-description": "You can fix them! The table below presents you with links to the statement on Wikidata and to the entry on the external source. You can compare the values and manually fix the mismatch, editing Wikidata or the external data source. After investigating and fixing the mismatch, you can indicate their status using the options from the drop-down. Consult the instructions",
+    "results-page-title": "What should I do with the mismatches?"
 }
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -25,6 +25,5 @@
     "column-external-value": "Value on external source",
     "column-upload-info": "Upload Info",
     "no-mismatches-found-message": "No mismatches have been found for:",
-    "results-page-description": "You can fix them! The table below presents you with links to the statement on Wikidata and to the entry on the external source. You can compare the values and manually fix the mismatch, editing Wikidata or the external data source. After investigating and fixing the mismatch, you can indicate their status using the options from the drop-down.",
     "results-page-title": "What should I do with the mismatches?"
 }
--- a/public/i18n/qqq.json
+++ b/public/i18n/qqq.json
@@ -21,6 +21,5 @@
     "column-wikidata-value": "Wikidata value column on mismatches table",
     "column-external-value": "External source column on mismatches table",
     "column-upload-info": "Upload Info column on mismatches table",
-    "no-mismatches-found-message": "A message when no mismatches have been found for the provided Item Ids",
-    "results-page-title": "The title of the Results page"
+    "no-mismatches-found-message": "A message when no mismatches have been found for the provided Item Ids"
 }
--- a/public/i18n/qqq.json
+++ b/public/i18n/qqq.json
@@ -20,5 +20,9 @@
     "column-property": "Property column on mismatches table",
     "column-wikidata-value": "Wikidata value column on mismatches table",
     "column-external-value": "External source column on mismatches table",
-    "column-upload-info": "Upload Info column on mismatches table"
+    "column-upload-info": "Upload Info column on mismatches table",
+    "no-mismatches-found-message": "A message when no mismatches have been found for the provided Item Ids",
+    "refine-items-selection": "The call to action on the Results Page to Return to the Homepage",
+    "results-page-description": "The description of the Results page",
+    "results-page-title": "The title of the Results page"
 }
--- a/public/i18n/qqq.json
+++ b/public/i18n/qqq.json
@@ -22,6 +22,5 @@
     "column-external-value": "External source column on mismatches table",
     "column-upload-info": "Upload Info column on mismatches table",
     "no-mismatches-found-message": "A message when no mismatches have been found for the provided Item Ids",
-    "results-page-description": "The description of the Results page",
     "results-page-title": "The title of the Results page"
 }
--- a/public/i18n/qqq.json
+++ b/public/i18n/qqq.json
@@ -22,7 +22,6 @@
     "column-external-value": "External source column on mismatches table",
     "column-upload-info": "Upload Info column on mismatches table",
     "no-mismatches-found-message": "A message when no mismatches have been found for the provided Item Ids",
-    "refine-items-selection": "The call to action on the Results Page to Return to the Homepage",
     "results-page-description": "The description of the Results page",
     "results-page-title": "The title of the Results page"
 }
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -1,12 +1,11 @@
 <template>
     <div class="page-container results-page">
         <Head title="Mismatch Finder - Results" />
-        <div v-if="Object.keys(results).length < item_ids.length" class="not-found">
-            <div id="top-button">
+        <div id="top-button">
                 <wikit-button
                     variant="normal"
-                    type="neutral"
                     native-type="button"
+                    type="neutral"
                     @click.native="returnHome"
                 >
                     &#x2190; {{ $i18n('refine-items-selection') }}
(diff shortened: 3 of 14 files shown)


resources/js/Pages/Results.vue
```diff
@@ -1,28 +1,51 @@
 <template>
     <div class="page-container results-page">
         <Head title="Mismatch Finder - Results" />
-        <section id="results" v-if="Object.keys(results).length">
-            <section class="item-mismatches" v-for="(mismatches, item, idx) in results" :key="idx">
+        <div v-if="Object.keys(results).length < item_ids.length" class="not-found">
```
itamargiv, line 4: Nice way to determine if the message should show!

resources/js/Pages/Results.vue
```diff
@@ -1,28 +1,51 @@
...
+                    native-type="button"
+                    @click.native="returnHome"
+                >
+                    &#x2190; {{ $i18n('refine-items-selection') }}
+                </wikit-button>
+            </div>
+            <section id="description-section">
+                <h2 class="h4">{{ $i18n('results-page-title') }}</h2>
+                <p id="about-description" >{{ $i18n('results-page-description') }}</p>
+            </section>
```
itamargiv, line 18: I don't think these UI elements are in scope for this task, they should be tackled at [T290861](https://phabricator.wikimedia.org/T290861). Either or, and since you already added the code, I think that they should also show up when we get full results.

resources/js/Pages/Results.vue
```diff
@@ -34,18 +57,31 @@
...
         props: {
             item_ids: Array as PropType<string[]>,
             results: Object as PropType<Result>,
             labels: Object as PropType<LabelMap>
         },
+        computed: {
+            labelsFromNotFound() {
+                const notFoundItems = this.item_ids.filter( element => !Object.keys( this.results ).includes(element) );
+                
+                const notFoundLabels = Object.entries(this.labels)
```
itamargiv, line 78: I don't think we should iterate over all the labels just to find the correct labels for missing items, can we just directly access the labels in the template? See suggestion below.

resources/js/Pages/Results.vue
```diff
@@ -1,28 +1,51 @@
...
+                <p id="about-description" >{{ $i18n('results-page-description') }}</p>
+            </section>
+            <section id="message-section">
+                <Message type="notice">
+                    <span>{{ $i18n('no-mismatches-found-message') }} 
+                        <span class="message-link" v-for="(label, item_id) in labelsFromNotFound" :key="item_id">
+                            <wikit-link 
+                                :href="`http://www.wikidata.org/wiki/${item_id}`">{{label}} ({{item_id}})
+                            </wikit-link>
+                        </span>
```
itamargiv, line 26: Instead of going through the entire labels object just to find the labels for a certain Id, we can just directly access it here, like so:

```suggestion
                        <span class="message-link" v-for="item_id in labelsFromNotFound" :key="item_id">
                            <wikit-link 
                                :href="`http://www.wikidata.org/wiki/${item_id}`">{{labels[item_id]}} ({{item_id}})
                            </wikit-link>
                        </span>
```

If you accept this change, we can also change the name of the function that returns the not found item ids.

resources/js/Pages/Results.vue
```diff
@@ -1,28 +1,51 @@
...
+                    &#x2190; {{ $i18n('refine-items-selection') }}
+                </wikit-button>
+            </div>
+            <section id="description-section">
+                <h2 class="h4">{{ $i18n('results-page-title') }}</h2>
+                <p id="about-description" >{{ $i18n('results-page-description') }}</p>
+            </section>
+            <section id="message-section">
+                <Message type="notice">
+                    <span>{{ $i18n('no-mismatches-found-message') }}
```
itamargiv, line 21: Why do we nest these spans within one another?

resources/js/Pages/Results.vue
```diff
@@ -56,15 +92,52 @@
                     value_label: this.labels[mismatch.wikidata_value] || null,
                     ...mismatch
                 }));
-            }
-        }
+            },
+            returnHome(): void {
+                this.$inertia.get( '/' );
+            },
```
itamargiv, line 98: Instead of forcing inertia back with JavaScript, I think it's best if we use an inertia link and the `as=wikit-button` attribute. See: https://inertiajs.com/links

itamargiv (changes requested):
Great work so far, there are a couple of changes that should be done, and I think we should probably also add a test to the `ResultsTest` browser test. But after that we are good to go.

resources/js/Pages/Results.vue
guergana: :D

resources/js/Pages/Results.vue
guergana: Applied suggestion.

resources/js/Pages/Results.vue
guergana: We'll address this in ticket [T290864](https://phabricator.wikimedia.org/T290864)

resources/js/Pages/Results.vue
guergana: Done.

resources/js/Pages/Results.vue
```diff
@@ -1,28 +1,50 @@
...
+                    @click.native="returnHome"
+                >
+                    &#x2190; {{ $i18n('refine-items-selection') }}
+                </wikit-button>
+            </div>
+            <section id="description-section">
+                <h2 class="h4">{{ $i18n('results-page-title') }}</h2>
+                <p id="about-description" >{{ $i18n('results-page-description') }}</p>
+            </section>
+        <div v-if="notFoundItemIds.length" class="not-found">
```
Silvan-WMDE, line 18: I think the `<div>` isn't needed here, as the `v-if` can go into the `<Message>` element, which will also keep the section structure intact, when the Message is not displayed. See [@itamargiv's comment](https://github.com/wmde/wikidata-mismatch-finder/pull/110#discussion_r713971650) on the Homepage error message.

resources/js/Pages/Results.vue
```diff
@@ -1,28 +1,50 @@
...
 
 <script lang="ts">
     import { PropType } from 'vue';
-
     import { Head } from '@inertiajs/inertia-vue';
-    import { Link as WikitLink } from '@wmde/wikit-vue-components';
-
+    import { 
+        Link as WikitLink,
+        Button as WikitButton,
```
Silvan-WMDE, line 46: T290864

resources/js/Pages/Results.vue
```diff
@@ -34,18 +56,25 @@
...
     }
-
+     
     export default defineComponent({
         components: {
             Head,
             MismatchesTable,
-            WikitLink
+            WikitLink,
+            WikitButton,
```
Silvan-WMDE, line 65: T290864

resources/js/Pages/Results.vue
Silvan-WMDE: yes, so it can be moved out of here :-)

resources/js/Pages/Results.vue
```diff
@@ -56,15 +85,52 @@
...
+        display: inline-block;
+    }
+
+    &::after {
+        content: ", ";
     }
+
+    &:last-child::after {
+        content: "";
+    }
```
Silvan-WMDE, line 116: wow, that's nice :-)

resources/js/Pages/Results.vue
```diff
@@ -56,15 +85,52 @@
...
+}
+
+#about-description {
+    max-width: 705px;
+    margin-top: 8px;
+}
```
Silvan-WMDE, line 122: T290861

resources/js/Pages/Results.vue
```diff
@@ -56,15 +85,52 @@
...
+
+#top-button {
+    margin-bottom: 15px;
+}
+
```
Silvan-WMDE, line 127: T290864

tests/Browser/Pages/HomePage.php
```diff
@@ -41,7 +41,7 @@ public function elements()
         return [
             '@form' => '#items-form',
             '@items-input' => '@form textarea',
-            '@not-found' => 'p.not-found'
+            '@not-found' => 'div.not-found'
```
Silvan-WMDE, line 44: that line is actually obsolete, isn't it? But I think it's unrelated to this task.

tests/Browser/ResultsTest.php
```diff
@@ -24,6 +24,73 @@ public function test_shows_item_ids()
...
+                    ->assertSee($mismatch->external_value)
+                    ->assertSeeLink($mismatch->importMeta->user->username)
+                    ->assertSee($mismatch->importMeta->created_at->toDateString());
+            }
+
+            $browser->assertSee('No mismatches have been found for')
+                ->assertSeeLink('Q1')
+                ->assertSeeLink('Q666');
+        });
+    }
```
Silvan-WMDE, line 92: This entire test basically copies the two tests that you already have: one for the notice message box and one for the correct result tables. Since these two are actually covered with all their details, I don't think that the same level of detail needs to be repeated. Instead we could simply test only for what we want to make sure: that both the message box and the results table are there in cases when mismatches for some items have been found, but not for others.

I think it can be done in a rather simple and easy to understand test:
```suggestion
    public function test_shows_table_for_existing_item_and_message_for_non_existing_item()
    {
        // add only one mismatch for Q1
        Mismatch::factory()
            ->for(ImportMeta::factory()->for(User::factory()->uploader()))
            ->state(['statement_guid' => 'Q1$a2b48f1f-426d-91b3-1e0e-1d3c7b236bd0'])
            ->create();

        $this->browse(function (Browser $browser) {
            // check results for Q1 and Q2
            $browser->visit(new ResultsPage('Q1|Q2'));

            $browser->assertSeeIn('.results', 'Q1');
            $browser->assertSeeIn('.wikit-Message--notice', 'Q2');
        });
    }
```

Silvan-WMDE (changes requested):
Nice work, thank you!
Below I marked all the parts that should be covered in the tasks for the intro text and the button and I made a suggestion on a much simpler test for the combined message/results case.

tests/Browser/Pages/HomePage.php
guergana: Yes, this is in main. It was introduced in a PR that has already been merged. I can remove it in this task. Should i?

resources/js/Pages/Results.vue
itamargiv: I would actually put it on the `section` element, so that it doesn't appear on the page at all if there are no messages

resources/js/Pages/Results.vue
itamargiv: Clever! :heart_eyes:

tests/Browser/ResultsTest.php
itamargiv: Oh wow, yeah that's more succinct!

tests/Browser/Pages/HomePage.php
Silvan-WMDE: Sure, why not. Actually, it is in fact related to this task, right? (even though it's the `HomePage` test)

guergana: Done.

public/i18n/en.json
```diff
@@ -23,5 +23,7 @@
     "column-property": "Property",
     "column-wikidata-value": "Value on Wikidata",
     "column-external-value": "Value on external source",
-    "column-upload-info": "Upload Info"
+    "column-upload-info": "Upload Info",
+    "no-mismatches-found-message": "No mismatches have been found for:",
+    "results-page-title": "What should I do with the mismatches?"
```
Silvan-WMDE, line 28:
```suggestion
    "no-mismatches-found-message": "No mismatches have been found for:"
```

public/i18n/qqq.json
```diff
@@ -20,5 +20,7 @@
     "column-property": "Property column on mismatches table",
     "column-wikidata-value": "Wikidata value column on mismatches table",
     "column-external-value": "External source column on mismatches table",
-    "column-upload-info": "Upload Info column on mismatches table"
+    "column-upload-info": "Upload Info column on mismatches table",
+    "no-mismatches-found-message": "A message when no mismatches have been found for the provided Item Ids",
+    "results-page-title": "The title of the Results page"
```
Silvan-WMDE, line 25:
```suggestion
    "no-mismatches-found-message": "A message when no mismatches have been found for the provided Item Ids"
```

resources/js/Pages/Results.vue
```diff
@@ -1,28 +1,33 @@
...
-        <section id="results" v-if="Object.keys(results).length">
-            <section class="item-mismatches" v-for="(mismatches, item, idx) in results" :key="idx">
+        <section id="message-section" v-if="notFoundItemIds.length">
+            <Message type="notice">
+                <span>{{ $i18n('no-mismatches-found-message') }}</span> 
+                    <span class="message-link" v-for="item_id in notFoundItemIds" :key="item_id">
+                        <wikit-link 
+                            :href="`http://www.wikidata.org/wiki/${item_id}`">{{labels[item_id]}} ({{item_id}})
+                        </wikit-link>
+                    </span>
```
Silvan-WMDE, line 11:
```suggestion
                <span>{{ $i18n('no-mismatches-found-message') }}</span> 
                <span class="message-link" v-for="item_id in notFoundItemIds" :key="item_id">
                    <wikit-link 
                        :href="`http://www.wikidata.org/wiki/${item_id}`">{{labels[item_id]}} ({{item_id}})
                    </wikit-link>
                </span>
```

Silvan-WMDE (commented):
minor fixes

resources/js/Pages/Results.vue
```diff
@@ -1,28 +1,33 @@
...
+                <span>{{ $i18n('no-mismatches-found-message') }}</span> 
+                <span class="message-link" v-for="item_id in notFoundItemIds" :key="item_id">
+                    <wikit-link 
+                        :href="`http://www.wikidata.org/wiki/${item_id}`">{{labels[item_id]}} ({{item_id}})
+                    </wikit-link>
+                </span>
+            </Message>
+        </section>
+        <section class="results" v-if="Object.keys(results).length">
+            <section v-for="(mismatches, item, idx) in results" :key="idx">
```
Silvan-WMDE, line 15:
```suggestion
        <section id="results" v-if="Object.keys(results).length">
            <section class="item-mismatches" v-for="(mismatches, item, idx) in results" :key="idx">
```

Silvan-WMDE (approved):
Done yay \o/ - Thanks a lot for dealing with all the request changes :+1:

itamargiv (approved):
:100: Awesome work, thank you!

PR #116, opened 2021-09-28
Fix width of error message box on home page

Silvan-WMDE:
This adjusts a possible error message box on the home page to have
the same width as the intro text above. The layout change now also
justifies moving the error message into the intro section, rather than
creating a second section with the same hard coded px size.

Bug: [T289344](https://phabricator.wikimedia.org/T289344)

### DIFF
--- a/resources/js/Pages/Home.vue
+++ b/resources/js/Pages/Home.vue
@@ -135,16 +135,13 @@
 <style lang="scss">
 @import '~@wmde/wikit-tokens/dist/_variables.scss';
 
-#about-description {
+#intro-section,
+#message-section {
     max-width: 705px;
 }
 
-#message-section {
-    max-width: 675px;
-
-    .wikit-Message {
-        border-radius: $border-radius-base;
-    }
+#message-section .wikit-Message {
+    border-radius: $border-radius-base;
 }
 
 #items-form {


resources/js/Pages/Home.vue
```diff
@@ -4,9 +4,7 @@
         <section id="intro-section">
             <h2 class="h4">{{ $i18n('about-mismatch-finder-title') }}</h2>
             <p id="about-description" >{{ $i18n('about-mismatch-finder-description') }}</p>
-        </section>
 
-        <section id="message-section">
```
itamargiv, line 9: We probably want to retain this in it's own section, both for the padding added by the design, but also for semantics.

resources/js/Pages/Home.vue
```diff
@@ -135,12 +133,8 @@
 <style lang="scss">
 @import '~@wmde/wikit-tokens/dist/_variables.scss';
 
-#about-description {
+#intro-section {
```
itamargiv, line 136: Nice retargeting! In order to reuse this rule for both sections I would do something like this:

```suggestion
#intro-section,
#message-section {
```

itamargiv (changes requested):
Great work, thanks! I left one comment below

resources/js/Pages/Home.vue
Silvan-WMDE: Do I leave the border-radius of .wikit-Message inside this shared rule then?
```
#about-description,
#message-section {
    max-width: 705px;

    .wikit-Message {
        border-radius: $border-radius-base;
}
```
or only share the 705px and define the radius alone, on the top level?
```
#about-description,
#message-section {
    max-width: 705px;
}

.wikit-Message {
    border-radius: $border-radius-base;
}
```

itamargiv: second option please :)

Silvan-WMDE: cool, thanks!

itamargiv (approved):
Great work!

PR #117, opened 2021-09-28
Replace table with wikit component

itamargiv:
This change replaces the local table component, with the up streamed WiKit version.

Bug: [T291079](https://phabricator.wikimedia.org/T291079)


Silvan-WMDE (approved):
Less code - and the table still works \o/

PR #118, opened 2021-09-29
Add review status column

itamargiv:
This change adds a column to each item's mismatches table, to indicate the review status of the mismatches. Additionally, since there were no tests enabled for the results page, this change adds some missing tests, in preparation for adding more tests in a followup to this PR.

This change does not yet include data binding between the status of the mismatches in the table, and the mismatches in the results pages' model after a new status is selected. These changes will  be included in a followup.

Bug: [T289557](https://phabricator.wikimedia.org/T289557)

### DIFF
--- a/resources/js/Components/MismatchesTable.vue
+++ b/resources/js/Components/MismatchesTable.vue
@@ -5,6 +5,7 @@
                 <th>{{$i18n('column-property')}}</th>
                 <th>{{$i18n('column-wikidata-value')}}</th>
                 <th>{{$i18n('column-external-value')}}</th>
+                <th>{{$i18n('column-review-status')}}</th>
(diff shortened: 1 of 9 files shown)


resources/js/Components/MismatchesTable.vue
```diff
@@ -32,3 +33,10 @@ export default Vue.extend({
     }
 });
 </script>
+
+<style lang="scss">
+    .column-review-status {
+        // Ensures that the dropdowns are evenly wide
+        min-width: 35%;
```
itamargiv, line 40:
```suggestion
        width: 35%;
```

Silvan-WMDE (approved):
cool, thanks :100:

PR #121, opened 2021-10-01
Handle server errors when storing mismatch reviews

Silvan-WMDE:
This change redirects any error response when saving mismatch reviews back
to the results page and adds an 'errors' flash message to the Inertia response.
This way the Results page can pick up the flash message and indicate a server
error with a Message component, triggered by a computed prop. Testers will be
able to fake/force a server error by sending a 'X-Mismatch-Finder-Error' header
with the 'mismatch-review' path as value.

The error message with its styles is moved into a new 'MessageBox' component,
so that it can be re-used in and between pages.

Bug: [T292055](https://phabricator.wikimedia.org/T292055)


resources/js/Components/MessageBox.vue
```diff
@@ -0,0 +1,31 @@
...
+            Message
+        }
+    });
+</script>
+
+<style lang="scss">
+@import '~@wmde/wikit-tokens/dist/_variables.scss';
+
+.wikit.wikit-Message {
+    max-width: 705px;
```
itamargiv, line 28: This is probably a bit too arbitrary here, and should be kept at the context of the page (to size the containing section rather than the message itself).

routes/web.php
```diff
@@ -36,6 +36,12 @@
     ->get('/results', [ResultsController::class, 'index'])
     ->name('results');
 
+Route::middleware('simulateError')
+    ->get('/mismatch-review', function () {
```
itamargiv, line 40: I don't think this is going to be a get request, also I don't think this is needed here at the moment. What I thought we would do is add an `update` method to the controller, and as such, we can also apply this middleware at the controller level.

itamargiv (changes requested):
Good job, I left one thing that might need to change in the component. But I agree with what you said in that chat, insofar as that it might not be needed after all.

app/Http/Middleware/SimulateError.php
```diff
@@ -24,10 +24,10 @@ class SimulateError
      */
     public function handle(Request $request, Closure $next)
     {
-        if ($request->header('X-Mismatch-Results-Error')) {
+        if ($request->header('X-Mismatch-Finder-Error') == $request->path()) {
```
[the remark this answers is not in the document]
Silvan-WMDE: Done: https://phabricator.wikimedia.org/T292282

resources/sass/app.scss
```diff
@@ -55,6 +55,16 @@ a {
...
 
+.wikit.wikit-Message {
+    border-radius: $border-radius-base;
+}
+
+#intro-section,
+#error-section,
+#message-section {
+        max-width: 705px;
+}
```
Silvan-WMDE, line 66: I'm not quite sure about the proper order here.

resources/js/Components/MessageBox.vue
Silvan-WMDE: Ok, I removed the MessageBox component and put styles for both the box radius and the message/intro sections width into `app.scss` now.

routes/web.php
Silvan-WMDE: moved into the ResultsController and changed to PUT. While it's not needed for functionality right now, the test with its simulated error still relies on this endpoint.

itamargiv (approved):
Looks great! Thank You!

PR #122, opened 2021-10-01
Update decisions at the results page level

itamargiv:
_**Note:** This change depends on #118_

This change ensures that user decisions regarding each mismatch are recorded at the result's page model. In order to skip a convoluted event trigger chain, an additional plugin was created to bubble events up from child to parent component.

[T289557](https://phabricator.wikimedia.org/T289557)

### DIFF
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -115,7 +115,6 @@
                 }));
             },
             recordDecision( decision: MismatchDecision ): void {
-                console.log(decision)
                 const itemDecisions = this.decisions[decision.item_id]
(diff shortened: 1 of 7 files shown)


resources/js/Pages/Results.vue
```diff
@@ -91,7 +114,15 @@
                     ...mismatch
                 }));
             },
-        },
+            recordDecision( decision: MismatchDecision ): void {
+                console.log(decision)
```
itamargiv, line 118:
```suggestion
```

Silvan-WMDE (approved):
:100: great stuff, thank you!

PR #125, opened 2021-10-05
Fix review log message test and log more data

Silvan-WMDE:
This change modifies the test for the review log message such, that it now creates a mismatch item in the database with an old timestamp and then asserts that the new timestamp of the update is logged to correctly. Also, this adds more mismatch data to the log message.

Bug: T289134

### DIFF
--- a/docs/AdminGuide.md
+++ b/docs/AdminGuide.md
@@ -48,3 +48,24 @@ Example User 4
 Example User 5
 Example User 6
 ```
+
+## Log Track of Users' Review Decisions
+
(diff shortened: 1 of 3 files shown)


docs/AdminGuide.md
```diff
@@ -48,3 +48,24 @@ Example User 4
 Example User 5
 Example User 6
 ```
+
+## Log Track of Users' Review Decisions
+
+Review decisions are written to mismatch entries in the database directly, without recording an edit history. Thus, for sanity reasons, a record of review decisions is kept on the filesystem in `storage/logs/mismatch_updates.log`.
```
guergana, line 54: Awesome!!! Thanks for adding docs!!!

PR #126, opened 2021-10-07
Test that a PUT request is sent

jakobw:


Silvan-WMDE (approved):
Looks good and works exactly as we wanted this to work. Thank you!

@guergana: What's left to check now is that `$inertia.put()` is not only called, but that it's called with the correct URL and payload.

PR #128, opened 2021-10-13
Combine endpoint for batch-updating mismatches with frontend form submission

guergana:
This PR adds a browser test to prove that the review status is sent after the "Apply changes" button is clicked.

Missing: Can't make the keyboard interaction work.

Bug: [T292058](https://phabricator.wikimedia.org/T292058)

### DIFF
--- a/resources/js/Components/MismatchesTable.vue
+++ b/resources/js/Components/MismatchesTable.vue
@@ -10,7 +10,11 @@
             </tr>
         </thead>
         <tbody>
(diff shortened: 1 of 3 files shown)


itamargiv (approved):
The initial issue with the test was that we needed to use a logged-in user to actually make these changes. 

This was fixed in my first commit. Additionally, I removed some redundant code and created a new testing component to make selecting options, and asserting values on the dropdown a little easier.

PR #159, opened 2021-10-20
Show only pending mismatches on results page

Silvan-WMDE:
This change modifies the database query for the results page such, that
only mismatches with review_status 'pending' are retrieved. This however
would lead to the undesired behaviour, that mismatches disappear from
the results page once their review_status has been modified and
submitted, since the inertia PUT request that has been used so far,
triggers a new database query and re-renders the page without the
updated mismatch entry.

Therefore, the PUT request to submit review decisions is now sent to the
endpoint as an asynchronous axios request in the background and returns
status code 200 upon success, instead of a redirect back to the results
page.

Additionally, the query is also modified to not return mismatches from expired
imports any longer and a feature test has been added.

Bug: [T293730](https://phabricator.wikimedia.org/T293730)

### DIFF
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -1,7 +1,7 @@
 <template>
     <div class="page-container results-page">
         <Head title="Mismatch Finder - Results" />
-        <section id="error-section" v-if="putError || unexpectedError">
+        <section id="error-section" v-if="unexpectedError">
             <Message type="error">{{ $i18n('server-error') }}</Message>
         </section>
         <section id="message-section">
@@ -84,8 +84,7 @@
     }
 
     interface ResultsState {
-        decisions: DecisionMap,
-        putError: string
+        decisions: DecisionMap
     }
 
     export default defineComponent({
@@ -125,8 +124,7 @@
         },
         data(): ResultsState {
             return {
-                decisions: {},
-                putError: ''
+                decisions: {}
             }
         },
         methods: {
@@ -152,12 +150,13 @@
                     // use axios in order to preserve saved mismatches
                     try {
                         await axios.put('/mismatch-review', this.decisions[item]);
+
+                        // remove decision from this.decisions after it has been
+                        // sent to the server successfully, to avoid sending them twice
+                        delete this.decisions[item];
                     } catch(e) {
-                        this.putError = item;
+                        console.error("saving review decisions has failed", e);
                     }
-                    // remove decision from this.decisions after it has been sent to the server to avoid sending
-                    // them twice
-                    delete this.decisions[item];
                 }
             },
         }
--- a/tests/Vue/Pages/Results.spec.js
+++ b/tests/Vue/Pages/Results.spec.js
@@ -189,6 +189,27 @@ describe('Results.vue', () => {
 
     });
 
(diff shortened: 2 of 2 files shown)


app/Http/Controllers/ResultsController.php
```diff
@@ -35,6 +36,10 @@ public function index(MismatchGetRequest $request, WikibaseAPIClient $wikidata):
 
         $mismatches = Mismatch::with('importMeta.user')
             ->whereIn('item_id', $itemIds)
+            ->where('review_status', 'pending')
+            ->whereHas('importMeta', function ($import) {
+                $import->where('expires', '>=', now());
+            })
```
itamargiv, line 42: Nice and concise!

resources/js/Pages/Results.vue
```diff
@@ -145,11 +145,11 @@
...
             },
-            send( item: string ): void {
+            async send( item: string ): Promise<void> {
 
-                if(this.decisions && Object.prototype.hasOwnProperty.call(this.decisions, item)){
-
-                    this.$inertia.put( '/mismatch-review', this.decisions[item] as unknown as RequestPayload );
+                if(this.decisions[item]){
+                    // use axios in order to preserve saved mismatches
+                    axios.put('/mismatch-review', this.decisions[item]);
```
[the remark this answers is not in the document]
itamargiv: yes, I would say so, otherwise this function will throw if the request doesn't succeed, and it will probably take the rest of the app with it. Also, it seems as though we forgot the `await` keyword, which means that right now, the decisions for the item will be deleted without necessarily completing the request successfully (async is lovely isn't it ;) )

tests/Browser/ResultsTest.php
```diff
@@ -157,8 +151,8 @@ public function test_apply_changes_button_submits_new_review_status()
             $browser->loginAs(User::factory()->create())
                 ->visit(new ResultsPage($mismatch->item_id))
                 ->within($dropdownComponent, function ($dropdown) {
-                    // make sure first value is displayed as it should
-                    $dropdown->assertOption('Mismatch on Wikidata')
+                    // make sure the item's initial review_status is 'pending'
+                    $dropdown->assertVue('value', null)
```
itamargiv, line 155: Nice! I forgot, that `pending` wasn't part of the `ReviewDecision` enum :scream:

tests/Vue/Pages/Results.spec.js
```diff
@@ -3,11 +3,16 @@ import Results from '@/Pages/Results.vue';
...
 import { ReviewDecision } from '@/types/Mismatch.ts';
+import axios from 'axios';
 
 // Stub the inertia vue components module entirely so that we don't run into
 // issues with the Head component.
 jest.mock('@inertiajs/inertia-vue', () => ({}));
 
+jest.mock("axios", () => ({
+    put: jest.fn()
+  }));
```
itamargiv, line 14:
```suggestion
}));
```
The nit has been picked XD

itamargiv (changes requested):
Great work, I like the new tests that were created, well done! There's just the one open issue of adding a `try... catch` block around the axios call, and making sure we `await` for it to complete.

PR #178, opened 2021-10-28
Confirmation Dialog

itamargiv:
This change implements the confirmation dialog that appears upon successful submission of review decisions. This does not yet include the option to ignore the dialog in the future, which will be addressed in a followup.

Bug: [T290953](https://phabricator.wikimedia.org/T290953)

### DIFF
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -55,7 +55,7 @@
                 namespace: 'next-steps-confirm'
             }]"
             @action="(_, dialog) => dialog.hide()"
-            dismissible
+            dismiss-button
         >
             <p>{{ $i18n('confirmation-dialog-message-intro') }}</p>
             <ul>
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -47,6 +47,23 @@
                 </form>
             </section>
         </section>
+        <wikit-dialog class="confirmation-dialog"
+            :title="$i18n('confirmation-dialog-title')"
+            ref="confirmation"
+            :actions="[{
+                label: $i18n('confirmation-dialog-button'),
+                namespace: 'next-steps-confirm'
+            }]"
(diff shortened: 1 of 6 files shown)


resources/js/Pages/Results.vue
```diff
@@ -150,6 +169,13 @@
                 };
             },
             async send( item: string ): Promise<void> {
+                // Casting to `any` since TS cannot understand $refs as
+                // component instances and complains about the usage of `show`
+                // See: https://github.com/vuejs/vue-class-component/issues/94
+                // Defaulting to any, as the alternative presents us with
+                // convoluted and unnecessary syntax.
+                const confirmationDialog = this.$refs.confirmation! as any;
```
guergana, line 177: Ohhh, typescript. Dont we all love it.

guergana (commented):
Looks good to me, tho I will let @Silvan-WMDE offer his input, since he always has interesting suggestions for tests.

Silvan-WMDE:
> Looks good to me, tho I will let @Silvan-WMDE offer his input, since he always has interesting suggestions for tests.

yes, one suggestion that I can think of is to add a vue test, which
a) makes sure the dialog is shown upon successful axios response
b) makes sure the dialog is **not** shown after a failed axios response

tests/Vue/Pages/Results.spec.js
```diff
@@ -231,4 +231,46 @@ describe('Results.vue', () => {
...
+                    decisions
+                }
+            }
+        });
+        const dialog = wrapper.find('.confirmation-dialog .wikit-Dialog');
+
+        await wrapper.vm.send(item_id);
+
+        expect(dialog.isVisible()).toBe(false);
+    });
```
Silvan-WMDE, line 275: Nice! I like these, and they do what they are supposed to do. Thanks!

Silvan-WMDE (approved):
Works and has tests - thank you! :100:

PR #181, opened 2021-10-29
Allow users to disable confirmation dialog

itamargiv:
_**Note:** This PR depends on #178_

This change allows the user to disable the confirmation dialog that appear when submitting decisions, and persists that state into the browser's local storage.

Additionally, some convenience methods were added to the results page testing object to apply and submit review decisions more easily within the browser tests.

Bug: [T290953](https://phabricator.wikimedia.org/T290953)


guergana (changes requested):
This is only working if I check the box and then reload the page. If I chose an option from the dropdown and the box opens and then proceed. And then I chose another option without reloading the page, the box appears again. I think this is due to the fact that you are only retrieving the value of the disableConfirmation in mounted and not saving it as a local data variable as well.

itamargiv:
> This is only working if I check the box and then reload the page. If I chose an option from the dropdown and the box opens and then proceed. And then I chose another option without reloading the page, the box appears again. I think this is due to the fact that you are only retrieving the value of the disableConfirmation in mounted and not saving it as a local data variable as well.

Thanks! It turned out that we just needed to ensure that the checkbox rest when the dialog is dismissed (rather than on every close, like before). I also made sure to fix the tests do that they would fail  in this case, as they should have.

Silvan-WMDE (approved):
Works locally, great work again - thank you! :tada:

guergana (approved):
beautiful

PR #184, opened 2021-11-01
Change width of review-status column to 25%

Silvan-WMDE:
**NOTE: This PR is based on PR #85, which should be merged first.**

The review status column should be 280px at an overall table width of
1120px, which is 25%.

Bug: [T294577](https://phabricator.wikimedia.org/T294577)


itamargiv (approved):
Wonderful!

PR #189, opened 2021-11-04
Add dialog with instructions to results intro text

Silvan-WMDE:
This change adds a quiet button to the Result Page's intro text,
which will open a dialog with detailed instructions about how to work
with the mismatches and how to resolve them properly.

Bug: [T290862](https://phabricator.wikimedia.org/T290862)

### DIFF
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -36,7 +36,7 @@
     "log-in-message": "Please <a href=\"$1\">log in</a> to be able to make any changes.",
     "results-page-description": "You can fix them! The table below presents you with links to the statement on Wikidata and to the entry on the external source. You can compare the values and manually fix the mismatch, editing Wikidata or the external data source. After investigating and fixing the mismatch, you can indicate their status using the options from the drop-down.",
     "results-page-title": "What should I do with the mismatches?",
-    "results-consult-instructions-link": "Consult the instructions",
+    "results-instructions-button": "Instructions",
     "confirmation-dialog-title": "Next steps",
     "results-read-full-description-link": "read the full description",
     "confirmation-dialog-message-intro": "Now that the changes are submitted successfully, remember:",
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -36,7 +36,7 @@
     "log-in-message": "Please <a href=\"$1\">log in</a> to be able to make any changes.",
     "results-page-description": "You can fix them! The table below presents you with links to the statement on Wikidata and to the entry on the external source. You can compare the values and manually fix the mismatch, editing Wikidata or the external data source. After investigating and fixing the mismatch, you can indicate their status using the options from the drop-down.",
     "results-page-title": "What should I do with the mismatches?",
-    "results-instructions-button": "Instructions",
+    "results-instructions-button": "Instruction",
     "confirmation-dialog-title": "Next steps",
     "results-read-full-description-link": "read the full description",
     "confirmation-dialog-message-intro": "Now that the changes are submitted successfully, remember:",
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -36,6 +36,7 @@
     "log-in-message": "Please <a href=\"$1\">log in</a> to be able to make any changes.",
     "results-page-description": "You can fix them! The table below presents you with links to the statement on Wikidata and to the entry on the external source. You can compare the values and manually fix the mismatch, editing Wikidata or the external data source. After investigating and fixing the mismatch, you can indicate their status using the options from the drop-down.",
     "results-page-title": "What should I do with the mismatches?",
+    "results-consult-instructions-link": "Consult the instructions",
     "confirmation-dialog-title": "Next steps",
     "results-read-full-description-link": "read the full description",
     "confirmation-dialog-message-intro": "Now that the changes are submitted successfully, remember:",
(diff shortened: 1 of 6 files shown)


resources/js/Pages/Results.vue
```diff
@@ -177,6 +214,12 @@
             }
         },
         methods: {
+            showInstructionsDialog(e: Event) {
+                e.preventDefault();
```
itamargiv, line 218: We probably don't need this anymore, since we are not using a link.

resources/js/Pages/Results.vue
```diff
@@ -267,4 +310,25 @@ h2 {
     margin-top: $dimension-layout-xsmall;
 }
 
+p.list-intro {
+    margin-bottom: 0
+}
+
+#description-section {
+    position:relative;
```
itamargiv, line 318: I'm not sure we need this relative positioning here.

resources/js/Pages/Results.vue
```diff
@@ -267,4 +310,25 @@ h2 {
...
+
+    .description-header {
+        display: flex;
+        flex-direction: row;
+        align-items: center;
```
itamargiv, line 323: In this case, we should also add `justify-content: space-between` which will ensure that each child of the header will take the most extreme side.

resources/js/Pages/Results.vue
```diff
@@ -267,4 +310,25 @@ h2 {
...
+
+        h2 {
+            width: 100%;
+        }
+
+        .instructions-button {
+            padding: 6px 13px;
+        }
```
itamargiv, line 331: I think we might not need these, if we employ `justify-content: space-between`

resources/sass/app.scss
```diff
@@ -102,7 +102,7 @@ li {
 #intro-section,
 #error-section,
 #message-section,
-#about-description {
+#description-section {
```
itamargiv, line 105: nice! thank you

tests/Vue/Pages/Results.spec.js
```diff
@@ -22,6 +22,25 @@ describe('Results.vue', () => {
...
+    it('displays intro text and instructions button', () => {
+        const wrapper = mount(Results, { mocks });
+
+        const intro = wrapper.find('#about-description');
+        expect(intro.isVisible()).toBe(true);
+
+        const instructionsButton = wrapper.find('.instructions-button');
+        expect(instructionsButton.isVisible()).toBe(true);
+
+    });
```
itamargiv, line 34: Usually we don't test for specific sections to be rendered, as this might make the tests too brittle (in case a class or id changes) which is really like testing function internals. This also doesn't really ensure us that these selectors actually contain the content we intend them to. So I think the test might be redundant (unusual for me to say, I know, and it also pains me to say so :confounded: )

tests/Vue/Pages/Results.spec.js
```diff
@@ -22,6 +22,25 @@ describe('Results.vue', () => {
...
+
+    it('shows dialog after clicking the instructions button', async () => {
+        const wrapper = mount(Results, { mocks });
+        await wrapper.find('.instructions-button').trigger('click');
```
itamargiv, line 38: Nice! In this case we are testing an interaction, which can be considered part of the "public interface" of the component (as clean as these things can be).

itamargiv (commented):
Fantastic work! And even more fantastic turnaround after a change of approach. Thank you! I just left a couple of comments about css and tests, otherwise, Looks good.

public/i18n/en.json
```diff
@@ -36,6 +36,7 @@
     "log-in-message": "Please <a href=\"$1\">log in</a> to be able to make any changes.",
     "results-page-description": "You can fix them! The table below presents you with links to the statement on Wikidata and to the entry on the external source. You can compare the values and manually fix the mismatch, editing Wikidata or the external data source. After investigating and fixing the mismatch, you can indicate their status using the options from the drop-down.",
     "results-page-title": "What should I do with the mismatches?",
+    "results-instructions-button": "Instructions",
```
guergana, line 39: I guess then this has to be changed to "Instruction".. :shrug:

resources/js/Pages/Results.vue
Silvan-WMDE: done

resources/js/Pages/Results.vue
Silvan-WMDE: That's right, I think that line remained from a previous, slightly different attempt to place the button.

resources/js/Pages/Results.vue
Silvan-WMDE: Yes, that's right, and the button padding isn't needed, either - we just use the component's dimensions.

Gone.

tests/Vue/Pages/Results.spec.js
Silvan-WMDE: This is actually a result of my "test first" approach on the task, but you're right, we can easily live without this one.

public/i18n/en.json
Silvan-WMDE: Changed. :man_shrugging: 

Isn't it funny that I use "instruction**s** button" and "instruction**s** dialog" everywhere in the code? :-)

guergana: Yes, I am almost sure there will be copy changes here.

guergana (approved):
Beautiful! Thank you.

PR #191, opened 2021-11-05
Results back button

itamargiv:
This change adds a back button to the results page, and ensures that the form state is maintained even after submitting item ids.

Bug: [T295011](https://phabricator.wikimedia.org/T295011)

### DIFF
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -53,5 +53,5 @@
     "instructions-dialog-message-instruction-external": "Mismatch on external data source: use this option if the error lies in the external data source. When possible, please inform the data source's maintainers of the error or correct the mismatch yourself.",
     "instructions-dialog-message-instruction-both": "Both are wrong: use this option if you find there are errors on both sides. When possible, please use the link to the Wikidata statement to correct the mismatch and/or inform the data source’s maintainers of the error.",
     "instructions-dialog-message-instruction-none": "None of the above: use this option if none of the other options are applicable. This might for example be the case if the mismatch has already been resolved by someone else.",
-    "results-back-button": "Refine Search Results"
+    "results-back-button": "Refine Item Selection"
 }
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -52,6 +52,6 @@
     "instructions-dialog-message-instruction-wikidata": "Mismatch on Wikidata: use this option if Wikidata's data is inaccurate. Please use the link to the Wikidata statement and edit it there to correct the mismatch.",
     "instructions-dialog-message-instruction-external": "Mismatch on external data source: use this option if the error lies in the external data source. When possible, please inform the data source's maintainers of the error or correct the mismatch yourself.",
     "instructions-dialog-message-instruction-both": "Both are wrong: use this option if you find there are errors on both sides. When possible, please use the link to the Wikidata statement to correct the mismatch and/or inform the data source’s maintainers of the error.",
-    "instructions-dialog-message-instruction-none": "None of the above: use this option if none of the other options are applicable. This might for example be the case if the mismatch has already been resolved by someone else."
-
+    "instructions-dialog-message-instruction-none": "None of the above: use this option if none of the other options are applicable. This might for example be the case if the mismatch has already been resolved by someone else.",
+    "results-back-button": "Refine Search Results"
 }
--- a/resources/js/Pages/Home.vue
+++ b/resources/js/Pages/Home.vue
@@ -114,7 +114,7 @@
                 }
 
                 this.$store.commit('saveSearchedIds', this.form.itemsInput);
-                this.$inertia.get( '/results?ids=' + this.serializeInput());
+                this.$inertia.get( '/results', { ids: this.serializeInput() } );
             },
         },
         computed: {
--- a/resources/js/Pages/Home.vue
+++ b/resources/js/Pages/Home.vue
@@ -123,21 +123,15 @@
                 return (flashMessages.errors && flashMessages.errors.unexpected);
             },
             // spread to combine with local computed props
-            ...mapState({
-                loading: 'loading',
-                lastSearchedIds: 'lastSearchedIds'
-            }),
+            ...mapState(['loading','lastSearchedIds']),
         },
         data(): HomeState {
             return {
                 form: {
-                    itemsInput: ''
+                    itemsInput: this.$store.state.lastSearchedIds
                 },
                 validationError: null
             }
-        },
-        mounted(){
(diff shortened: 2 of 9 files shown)


resources/js/app.ts
```diff
@@ -26,14 +26,18 @@ function createStore(): Store<{loading: boolean}>{
 
     const store = new Store({
         state: {
-            loading: false
+            loading: false,
+            lastSearchedIds: ''
```
Silvan-WMDE, line 30: hm... I see...

tests/Vue/Pages/Home.spec.js
```diff
@@ -18,21 +18,17 @@ describe('Home.vue', () => {
     const localVue = createLocalVue();
     localVue.use(Vuex);
 
-    const store = new Vuex.Store();
-
     it('sanitises input with empty lines', async () => {
-        const formData = { itemsInput: 'Q1\n\nQ2\n' };
+
+        const store = new Vuex.Store({state: { lastSearchedIds: 'Q1\n\nQ2\n' }});
```
Silvan-WMDE, line 23: Ah, here the unconditional replacement of the `form.itemsInput` bites you, right? :-)

I think the two features should not be mixed in a single test. One should be testing that input is sanitised and a second one should make sure that the input is restored when `lastSearchedIds` is passed via the Store. In the first case, when mounting the wrapper with an empty store, I guess you will now have to set `form.itemsInput` directly.

tests/Vue/Pages/Home.spec.js
```diff
@@ -18,21 +18,17 @@ describe('Home.vue', () => {
...
             localVue,
-            store,
-            data() {
-                return { form: formData };
-            },
+            store
         });
 
-        expect( wrapper.vm.serializeInput() ).toEqual('Q1|Q2');
+        expect( wrapper.vm.serializeInputUrl() ).toEqual('Q1|Q2');
```
Silvan-WMDE, line 31: This can probably remain if you remove the new `serializeInputText` method. The name also wouldn't make much sense, since it's not the URL that is serialized, but only the items ids.

Silvan-WMDE (changes requested):
Works great and looks like a really clean solution. Pls find just a few suggestions how to do it with less code. :-)

resources/js/Pages/Home.vue
```diff
@@ -133,6 +138,9 @@
                 },
                 validationError: null
             }
+        },
+        mounted(){
+            this.form.itemsInput = this.$store.state.lastSearchedIds;
```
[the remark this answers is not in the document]
guergana: lastSearchedIds is never null. It is initialized as an empty string. see app.ts

resources/js/app.ts
guergana: yes

resources/js/Pages/Home.vue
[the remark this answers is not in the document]
itamargiv: Do we really need this on `mounted()`? Shouldn't `mapState` take care of mapping from the global store into our local form state?

resources/js/Pages/Home.vue
```diff
@@ -79,9 +79,12 @@
                 // see: https://stackoverflow.com/a/281335/1619792
                 return this.splitInput().filter(x => x);
             },
-            serializeInput: function(): string {
+            serializeInputUrl: function(): string {
                 return this.sanitizeArray().join('|');
             },
+            serializeInputText: function(): string {
+                return this.splitInput().join('\n');
```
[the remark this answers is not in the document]
itamargiv: Perhaps, depending on what is answered in my question below.

tests/Vue/Pages/Home.spec.js
guergana: Adressed with the removal of the serializeInputText commit

public/i18n/en.json
```diff
@@ -52,6 +52,6 @@
...
     "instructions-dialog-message-instruction-external": "Mismatch on external data source: use this option if the error lies in the external data source. When possible, please inform the data source's maintainers of the error or correct the mismatch yourself.",
     "instructions-dialog-message-instruction-both": "Both are wrong: use this option if you find there are errors on both sides. When possible, please use the link to the Wikidata statement to correct the mismatch and/or inform the data source’s maintainers of the error.",
-    "instructions-dialog-message-instruction-none": "None of the above: use this option if none of the other options are applicable. This might for example be the case if the mismatch has already been resolved by someone else."
-
+    "instructions-dialog-message-instruction-none": "None of the above: use this option if none of the other options are applicable. This might for example be the case if the mismatch has already been resolved by someone else.",
+    "results-back-button": "Refine Search Results"
```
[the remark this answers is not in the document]
guergana: ok. this was added by Itamar before. I didn't really check the copy. Thanks.

resources/js/Pages/Home.vue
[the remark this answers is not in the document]
guergana: yes, we don't :D added to data instead

tests/Vue/Pages/Home.spec.js
guergana: Two tests added. Good observation.

resources/js/Pages/Results.vue
```diff
@@ -1,6 +1,9 @@
 <template>
     <div class="page-container results-page">
         <Head title="Mismatch Finder - Results" />
+        <wikit-button class="back-button" @click.native="() => $inertia.visit('/')">
```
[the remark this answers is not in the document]
guergana: Gotcha. Replaced

tests/Vue/Pages/Home.spec.js
```diff
@@ -18,21 +18,39 @@ describe('Home.vue', () => {
...
+        const store = new Vuex.Store({state: {lastSearchedIds: 'Q4\nQ55'} });
+
+        const wrapper = mount(Home, { 
+            mocks,
+            localVue,
+            store
+        });
+
+        expect( wrapper.vm.form.itemsInput ).toEqual('Q4\nQ55');
+    });
```
Silvan-WMDE, line 54:
```suggestion
    it('restores lastSearchIds value from store on page load', async () => {

        const itemsInput = 'Q1\n\nQ2\n';
        const store = new Vuex.Store({state: {lastSearchedIds: itemsInput} });

        const wrapper = mount(Home, { 
            mocks,
            localVue,
            store
        });

        expect( wrapper.vm.form.itemsInput ).toEqual(itemsInput);
    });
```

Silvan-WMDE (approved):
let's get this merged :100:

PR #192, opened 2021-11-05
Open external links in a new tab

guergana:
Bug: [T292403](https://phabricator.wikimedia.org/T292403)

### DIFF
--- a/resources/js/Components/MismatchRow.vue
+++ b/resources/js/Components/MismatchRow.vue
@@ -3,12 +3,13 @@
         <td :data-header="$i18n('column-property')">
             <wikit-link
               class="break-line-link"
+              target="blank"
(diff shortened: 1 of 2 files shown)


resources/js/Components/MismatchRow.vue
```diff
@@ -3,12 +3,13 @@
         <td :data-header="$i18n('column-property')">
             <wikit-link
               class="break-line-link"
+              target="blank"
```
Silvan-WMDE, line 6:
```suggestion
              target="_blank"
```

Silvan-WMDE (changes requested):
Found something! :laughing:

resources/js/Components/MismatchRow.vue
guergana: oh, yes, i had the feeling i had that made that mistake.

PR #195, opened 2021-11-08
Bugfix: Remove empty lines from input

guergana:
Bug: [T292410](https://phabricator.wikimedia.org/T292410)

### DIFF
--- a/resources/js/Pages/Home.vue
+++ b/resources/js/Pages/Home.vue
@@ -75,6 +75,8 @@
                 return this.form.itemsInput.split( '\n' );
             },
             sanitizeArray: function(): Array<string> {
+                // this filter function removes all falsy values
+                // see: https://stackoverflow.com/a/281335/1619792
                 return this.splitInput().filter(x => x);
             },
             serializeInput: function(): string {
--- a/resources/js/Pages/Home.vue
+++ b/resources/js/Pages/Home.vue
@@ -74,8 +74,11 @@
             splitInput: function(): Array<string> {
                 return this.form.itemsInput.split( '\n' );
             },
+            sanitizeArray: function(): Array<string> {
+                return this.splitInput().filter(x => x);
+            },
             serializeInput: function(): string {
-                return this.splitInput().join('|');
+                return this.sanitizeArray().join('|');
             },
             checkEmpty(): void {
                 if( !this.form.itemsInput ) {
@@ -89,9 +92,9 @@
                 this.validationError = null;
                 this.checkEmpty();
 
-                let valid = this.splitInput().every( function( currentValue: string ) {
+                let valid = this.sanitizeArray().every( function( currentValue: string ) {
                     let trimmedLine = currentValue.trim();
-                    return trimmedLine == '' || trimmedLine.match( /^[Qq]\d+$/ );
+                    return trimmedLine.match( /^[Qq]\d+$/ );
                 });
 
                 if( !valid ) {
--- a/tests/Browser/ItemsFormTest.php
+++ b/tests/Browser/ItemsFormTest.php
@@ -74,4 +74,16 @@ class ItemsFormTest extends DuskTestCase
             $this->assertStringContainsString('--error', $browser->attribute('@items-input', 'class'));
         });
     }
+
+    public function test_sends_sanitized_input_when_given_empty_line_breaks()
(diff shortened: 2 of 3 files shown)


tests/Vue/Pages/Home.spec.js
```diff
@@ -0,0 +1,74 @@
...
+        // window.location = {
+        //     ...window.location,
+        //     reload: jest.fn()
+        // }
+
+        //window.location.assign = jest.fn();
+        await wrapper.find('.wikit-Button[type="submit"]').trigger('click');
+        //expect(window.location.origin).toHaveBeenCalledWith('/results?ids=Q1');
+        //expect(wrapper.vm.serializeInput)
+        expect( inertiaGet ).toHaveBeenCalledWith('/results?ids=Q1|Q2');
```
Silvan-WMDE, line 70: I don't even think you have to test the full `send()` method. Why not limit the test to the sanitisation function you added and make sure that `{ itemsInput: 'Q1\n\nQ2\n' }` will be transformed to `'Q1|Q2'` (and not 'Q1||Q2|') by `serializeInput()`?

Silvan-WMDE: I added a suggested test as a [PR to yours](https://github.com/wmde/wikidata-mismatch-finder/pull/197) - pls feel free to use it as it is or adjust it if you think more or different tests are needed.

guergana: Thank you.

Silvan-WMDE (changes requested):
Nice and easy solution - works great! Just add a comment on the filter method pls, because the internals are actually not immediately obvious.

Also I'd like to hear @itamargiv's thoughts on the double tests: are both justified or should we pick one? If so, which one?

resources/js/Pages/Home.vue
```diff
@@ -74,8 +74,11 @@
             splitInput: function(): Array<string> {
                 return this.form.itemsInput.split( '\n' );
             },
+            sanitizeArray: function(): Array<string> {
+                return this.splitInput().filter(x => x);
```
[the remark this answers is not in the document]
itamargiv: Ah, this is a clever way to do it, nicely done @guergana! 

I think we can also use a more explicit notation here, it might be more code, but I think it could be worth it. Otherwise, this will also filter out values like `0` (which are not allowed anyhow due to validation, but perhaps it's just best to make it more explicit)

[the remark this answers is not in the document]
guergana: Ah, yes, my bad, I forgot.

resources/js/Pages/Home.vue
```diff
@@ -89,9 +92,9 @@
                 this.validationError = null;
                 this.checkEmpty();
 
-                let valid = this.splitInput().every( function( currentValue: string ) {
+                let valid = this.sanitizeArray().every( function( currentValue: string ) {
                     let trimmedLine = currentValue.trim();
-                    return trimmedLine == '' || trimmedLine.match( /^[Qq]\d+$/ );
```
itamargiv, line 94: Well caught! this was indeed the source of the bug.

tests/Browser/ItemsFormTest.php
```diff
@@ -74,4 +74,16 @@ public function test_invalid_item_list_yields_error()
             $this->assertStringContainsString('--error', $browser->attribute('@items-input', 'class'));
         });
     }
+
+    public function test_sends_sanitized_input_when_given_empty_line_breaks()
```
itamargiv, line 78: Nice tests! I really like that we added both a unit test and a browser test.

resources/js/Pages/Home.vue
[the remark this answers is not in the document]
guergana: It's done. :)

itamargiv (changes requested):
Looks great to me, thank you! Just one small change to make the filtering logic more explicit, otherwise, can be merged.

itamargiv (approved):

resources/js/Pages/Home.vue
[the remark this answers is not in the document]
Silvan-WMDE: coolio

Silvan-WMDE (approved):
:clap:

PR #196, opened 2021-11-08
Use button instead of link for description dialog

Silvan-WMDE:
This change replaces the existing `<wikit-link>` element from the Results
Page's "Upload Info" column with a `<wikit-button>` component that is
styled in such a way, that it appears as an inline element.

Bug: [T290849](https://phabricator.wikimedia.org/T290849)

### DIFF
--- a/resources/js/Components/MismatchRow.vue
+++ b/resources/js/Components/MismatchRow.vue
@@ -57,11 +57,16 @@
                 <span class="upload-date">{{uploadDate}}</span>
                 <div class="description">
                   {{uploadInfoDescription}}
-                  <wikit-link v-if="shouldTruncate" class="full-description-link" href="#" @click.native="showDialog">
-                    {{$i18n('results-read-full-description-link')}}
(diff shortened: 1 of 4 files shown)


resources/js/Components/MismatchRow.vue
```diff
@@ -181,7 +187,8 @@ export default Vue.extend({
     .wikit-Link__content {
       word-break: break-word;
     }
-    .wikit-Link.full-description-link {
-      display: inline;
+    .wikit-Button.full-description-button {
+      padding: 0px 2px;
```
guergana, line 191: I am guessing the look of the button has already been discussed.

guergana (approved):
Looks great. Thanks

resources/js/Components/MismatchRow.vue
Silvan-WMDE: Yes, it has. I'll merge it now, just to get it onto prod - then it can be discussed further. :-)

PR #199, opened 2021-11-10
Add arrow icon to button

guergana:
Can be tested by adding a rtl language json file to the `/i18n` folder and using the `uselang=<language>` flag. Example: `http://localhost:3000/results?ids=Q3|Q4&uselang=en`

Bug: [T294940](https://phabricator.wikimedia.org/T294940)

### DIFF
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -2,6 +2,9 @@
     <div class="page-container results-page">
         <Head title="Mismatch Finder - Results" />
         <wikit-button class="back-button" @click.native="() => $inertia.get('/', {} ,{ replace: true })">
+            <template #prefix>
+                <icon type="arrowprevious" size="medium" color="inherit" dir="ltr"/>
+            </template>
             {{ $i18n('results-back-button') }}
         </wikit-button>
         <section id="description-section">
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -3,7 +3,7 @@
         <Head title="Mismatch Finder - Results" />
         <wikit-button class="back-button" @click.native="() => $inertia.get('/', {} ,{ replace: true })">
             <template #prefix>
-                <icon type="arrowprevious" size="medium" color="inherit" :dir="iconDirection"/>
+                <icon type="arrowprevious" size="medium" color="inherit" :dir="pageDirection"/>
             </template>
             {{ $i18n('results-back-button') }}
         </wikit-button>
@@ -158,7 +158,8 @@
 
     interface ResultsState {
         decisions: DecisionMap,
-        disableConfirmation: boolean
+        disableConfirmation: boolean,
+        pageDirection: string
     }
 
     export default defineComponent({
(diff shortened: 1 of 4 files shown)


resources/js/Pages/Results.vue
```diff
@@ -2,6 +2,9 @@
     <div class="page-container results-page">
         <Head title="Mismatch Finder - Results" />
         <wikit-button class="back-button" @click.native="() => $inertia.visit('/')">
+            <template #prefix>
+                <icon type="arrowprevious" size="medium" color="inherit" dir="ltr"/>
```
itamargiv, line 6: Here we can derive the language from the body, we just need to add `dir="auto"` to the vody tag and use `window.getComputedStyles(document.body).dir`. See Proof of concept at: https://codepen.io/xumium/pen/PoKQGQL

itamargiv (changes requested):

resources/js/Pages/Results.vue
guergana: Good one!! Tho I think it should be in the `<html>` tag, not in the body one. See [this link](https://www.w3.org/International/questions/qa-html-dir)

resources/js/Pages/Results.vue
```diff
@@ -194,6 +197,9 @@
             unexpectedError() {
                 const flashMessages = this.$page.props.flash as FlashMessages;
                 return (flashMessages.errors && flashMessages.errors.unexpected);
+            },
+            iconDirection() {
+                return document.querySelector('html')?.getAttribute('dir');
```
itamargiv, line 202: So here we need to actually get the computed style rather that what the attribute was set to. It can be done with `window.getComputedStyle(document.documentRoot)`

resources/views/app.blade.php
```diff
@@ -1,12 +1,12 @@
 <!DOCTYPE html>
-<html lang="{{ App::currentLocale() }}">
+<html lang="{{ App::currentLocale() }}" dir="auto">
```
itamargiv, line 2: The reason that we will be setting this in body rather than HTML is because the "auto" value looks into the text content of the element to determine the directionality. See https://www.w3.org/International/questions/qa-html-dir#dirauto

itamargiv (changes requested):
A couple of small changes, looks great otherwise, thank you!

resources/js/Pages/Results.vue
guergana: `window.getComputedStyle(document.body).dir` gives typescript errors.

guergana: Cool that you could figure it out :)

itamargiv (approved):

resources/js/Pages/Results.vue
```diff
@@ -197,6 +201,7 @@
             }
         },
         mounted(){
+            this.pageDirection = window.getComputedStyle(document.body).direction;
```
guergana, line 204: nice work, @itamargiv Thanks.

Silvan-WMDE (approved):
works automagically - love it! :heart_eyes_cat:

PR #202, opened 2021-11-11
Remove casting to string of importmeta description

guergana:
Casting $importmeta->description to string when the value is null was causing the test to fail. 

Bug: [T295492](https://phabricator.wikimedia.org/T295492)


Silvan-WMDE (approved):
Ok, this changed test has passed twice in a row, so at least it doesn't break anything and hopefully we won't see it fail any more.

Thank you very much for fixing this! :pray:

PR #203, opened 2021-11-12
Add GitHub Action config for staging deployment

Silvan-WMDE:
This adds a GitHub action to deploy Mismatch Finder to the staging
environment on toolforge. Very similarly to the production config, it
first uploads the entire repo directory to a user's toolforge account,
and then syncs to the tool's account from there. The difference to
production is the tool being used ('mismatch-finder-staging') and the
staging directory name in the user's home directory, which is
'mismatch-finder-repo-next' here, instead of 'mismatch-finder-repo',
which is the one for production.

Staging deployments will be run on every push to main, as well as on
branches prefixed by 'staging' or 'gh-actions'.

Bug: T295585

### DIFF
--- a/.github/workflows/deploy-app-staging.yml
+++ b/.github/workflows/deploy-app-staging.yml
@@ -1,9 +1,10 @@
-name: Build and Deploy to production on Toolforge
+name: Build and Deploy to staging on Toolforge
 
 on:
   push:
    branches:
     - main
+    - staging/**
     - gh-actions/**
 
 jobs:
@@ -29,7 +30,7 @@ jobs:
       run: "npm run prod"
     - name: Remove node_modules before syncing
       run: "rm -rf node_modules"
-    - name: Syncing code to Toolforge-Production
+    - name: Syncing code to Toolforge-Staging
       uses: appleboy/scp-action@master
       with:
         host: ${{ secrets.HOST }}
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-app.yml
```diff
@@ -52,6 +52,6 @@ jobs:
           become mismatch-finder chmod -R g+rwx ~tools.mismatch-finder/mismatch-finder-repo
           # Change group of ~/mismatch-finder-repo (including symlinks) to tools.mismatch-finder
           chgrp --no-dereference -R tools.mismatch-finder ~/mismatch-finder-repo
-          rsync -rlgD --delete --delay-updates --exclude '.nfs*' --exclude mismatch-finder-repo/.env --exclude mismatch-finder-repo/storage/app/ ~/mismatch-finder-repo ~tools.mismatch-finder/
+          rsync -rlgD --delete --delay-updates --exclude '.nfs*' --exclude .env --exclude storage/app/ ~/mismatch-finder-repo ~tools.mismatch-finder/
```
Silvan-WMDE, line 55: this does the same thing, but is much shorter (esp in the staging version)

itamargiv (approved):
NICE! Tested a deployment with `gh-actions/*` prefix. It worked like a charm, no explosions woohoo :confetti_ball:

.github/workflows/deploy-app.yml
itamargiv: thank you, it is!

PR #205, opened 2021-11-15
Restore item input from Results URL

Silvan-WMDE:
This fixes an issue raised at: https://phabricator.wikimedia.org/T295011#7503295.

Additionally, since we now utilize the vuex store in the 'results' page as well, unit tests were adapted to ensure that the store is mocked in every test.

Bug: T295011

### DIFF
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -201,6 +201,10 @@
             }
         },
         mounted(){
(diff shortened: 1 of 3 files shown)


Silvan-WMDE (commented):
Great fix for the vue tests - thank you!

itamargiv (approved):
Nice quick thinking, thank you! :pray:

PR #206, opened 2021-11-15
Add staging deployment file

itamargiv:
This change introduces a new deployment file to be executed in staging environments, in order to run a continuous job for the queue runner in our staging environment.

Bug: [T290058](https://phabricator.wikimedia.org/T290058)


Silvan-WMDE (commented):
Thanks for adding the staging depoyment file, don't know why I missed that. :pray: 

However, I think we should give both `deployment.yaml` and `staging.yaml` proper names that somewhat tell what they are for.

Silvan-WMDE (approved):
worked nicely on [gh-actions/staging-continous-job](https://github.com/wmde/wikidata-mismatch-finder/tree/gh-actions/staging-continous-job) :+1:

PR #212, opened 2021-11-18
Add artisan commands to show and drop uploads

Silvan-WMDE:
This change adds artisan commands for admins to show the list of all mismatch uploads and to drop an entire upload by ID.

Bug: [T285845](https://phabricator.wikimedia.org/T285845)


app/Console/Commands/DropUpload.php
```diff
@@ -0,0 +1,71 @@
...
+        $mismatchesToDrop = Mismatch::where('import_id', '=', $import->id)->count();
+        $this->line(__(
+            'admin.dropUpload:dropping',
+            [
+                'id' => $import->id,
+                'mismatches' => $mismatchesToDrop
+            ]
+        ));
+
+        if ($this->confirm('Are you sure?')) {
```
itamargiv, line 58: I think we should probably put this message where we keep all our messages.

app/Console/Commands/DropUpload.php
```diff
@@ -0,0 +1,71 @@
...
+            Mismatch::where('import_id', '=', $import->id)->delete();
+            $import->delete();
```
itamargiv, line 60: I think it will be best if we put these deletions inside a [database transaction](https://laravel.com/docs/8.x/database#database-transactions). That way, the db will roll back in case something catastrophic happens while deleting.

app/Console/Commands/ShowUploads.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Models\ImportMeta;
+use App\Models\Mismatch;
+use Illuminate\Console\Command;
+
+class ShowUploads extends Command
```
itamargiv, line 9: Nice touch! Pretty great tool for completion!

docs/AdminGuide.md
```diff
@@ -49,6 +49,45 @@ Example User 5
...
+
+The following custom commands are provided to show and drop mismatch uploads:
+ * `php artisan uploads:show` will show you the list of all mismatch uploads whose import status is _completed_.
+ * `php artisan uploads:drop {id}` will delete an entire upload and all of its associated mismatches from Mismatch Finder's store.
+
+Instructions on how to use the commands can be displayed using:
+ * `php artisan help uploads:show` and
+ * `php artisan help uploads:drop`
+
+ **ATTENTION: Dropping an upload from the store will delete all its asociated mismatches, whether they have been reviewed or not.**
```
itamargiv, line 66: English docs convention usually puts these under "IMPORTANT" rather than "Attention".

docs/AdminGuide.md
```diff
@@ -49,6 +49,45 @@ Example User 5
...
+
+**Examples**
+
+```bash
```
itamargiv, line 70: Fantastic example! thank you!

itamargiv (changes requested):
Looks great, works great! Thank you very much. 

One thing I think we should insist on, though, is a clear domain language. In this case we are saying we are dropping and listing an **Upload**, but everywhere else we call this domain model an **Import**. In my opinion, and for the sake of clarity, we should call it an import here as well.

Otherwise, I left a couple of small comments, and then I think we're good to go.

app/Console/Commands/DropUpload.php
Silvan-WMDE: Yes, done. I didn't add any explicit error handling, I guess in case of catastrophy the command will just throw and exit with an error? Or does the success message belong in the `DB::transaction` block?

itamargiv: It'll throw, so I think it's okay

itamargiv (approved):
Nicely done! Thank You!

PR #213, opened 2021-11-18
Bugfix: Serve swagger-ui files from Toolforge CDN to avoid CSP issues

guergana:
Bug: [T287948](https://phabricator.wikimedia.org/T287948)


public/api-docs/index.html
```diff
@@ -4,8 +4,8 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
-    <script src="https://unpkg.com/swagger-ui-dist@3/swagger-ui-bundle.js" charset="UTF-8"></script>
-    <link rel="stylesheet" href="//unpkg.com/swagger-ui-dist@3/swagger-ui.css" />
+    <script src="./swagger-ui-dist@3/swagger-ui-bundle.js" charset="UTF-8"></script>
```
itamargiv, line 7: I would still ike to avoid including 3rd party code in our repo this way, since it will be hard to ensure it remains secure. However, I would also like to avoid having to add it as a dependency to our package json, etc.

So I snooped around the toolforge docs and I found this: 
https://wikitech.wikimedia.org/wiki/Help:Toolforge/Web#Load_external_assets_using_our_CDN_services

Looking at the available libraries at: 
https://cdnjs.toolforge.org/

It seems like `swagger-ui` is available there, so we can use that CDN mirror instead of `unpkg` and avoid violating the CSP header.

itamargiv (changes requested):
I think we can still avoid adding these files to our repo, see what I found in the comment.

public/api-docs/index.html
guergana: Ohhh, very nice. Will try that

guergana: Yeii. no console errors!!! https://mismatch-finder-staging.toolforge.org/api-docs/index.html

itamargiv (approved):
Nice! Thank you!

PR #218, opened 2021-11-19
Combine Dependabot updates

itamargiv:
This change combines several Dependabot updates into one branch.


Silvan-WMDE (approved):
go for it

PR #219, opened 2021-11-19
Add instructions on mismatch reviews via REST API

itamargiv:
This change adds instructions on how to review Mismatches using our REST API. In addition, we enable deep linking within our API documentation, to allow linking to particular endpoint specifications.

Bug: [T290182](https://phabricator.wikimedia.org/T290182)

### DIFF
--- a/public/api-docs/index.html
+++ b/public/api-docs/index.html
@@ -15,6 +15,7 @@
           SwaggerUIBundle({
             url: "/api-docs/mismatchfinder-api-spec.yml",
             dom_id: '#swagger-ui',
+            deepLinking: true,
             presets: [
               SwaggerUIBundle.presets.apis
             ]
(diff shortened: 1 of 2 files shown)


public/api-docs/index.html
```diff
@@ -15,6 +15,7 @@
           SwaggerUIBundle({
             url: "/api-docs/mismatchfinder-api-spec.yml",
             dom_id: '#swagger-ui',
+            deepLinking: true,
```
Silvan-WMDE, line 18: nice!

Silvan-WMDE (changes requested):
Thanks, great docs, I like it. Too bad we didn't include documentation when we actually implemented the features.

Just a few minor changes below.

Silvan-WMDE (approved):
that was quick :heavy_check_mark:

PR #223, opened 2021-11-22
Disable SimulateError middleware in production env

Silvan-WMDE:
Bug: [T292282](https://phabricator.wikimedia.org/T292282)

### DIFF
--- a/app/Http/Middleware/SimulateError.php
+++ b/app/Http/Middleware/SimulateError.php
@@ -7,11 +7,12 @@ use Illuminate\Http\Request;
 use RuntimeException;
 
 /**
- * This middleware will force a RuntimeException for testing purposes, if the
- * 'X-Mismatch-Finder-Error' HTTP header field is sent with the request and
- * its value matches the request path. It allows for simple feature tests of
- * the error routing behaviour as well as easy verification by testers, using
- * a browser extension such as https://addons.mozilla.org/de/firefox/addon/modheader-firefox/
+ * In non-production environments, this middleware will force a RuntimeException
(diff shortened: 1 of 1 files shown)


app/Http/Middleware/SimulateError.php
```diff
@@ -24,10 +25,12 @@ class SimulateError
...
-        if ($request->header('X-Mismatch-Finder-Error') == $request->path()) {
-            throw new RuntimeException("Simulated Server Error");
-        } elseif ($request->header('X-Mismatch-Finder-Not-Found') == $request->path()) {
-            abort(404);
+        if (!app()->environment('production')) {
+            if ($request->header('X-Mismatch-Finder-Error') == $request->path()) {
+                throw new RuntimeException("Simulated Server Error");
+            } elseif ($request->header('X-Mismatch-Finder-Not-Found') == $request->path()) {
+                abort(404);
+            }
```
itamargiv, line 33: Nice! A small style nitpick, in this case, I'd personally prefer to use a guard rather than wrap conditionals with conditionals. So the suggestion would be as follows:

```suggestion
        if (app()->environment('production')) {
            return;
        }
        
        if ($request->header('X-Mismatch-Finder-Error') == $request->path()) {
            throw new RuntimeException("Simulated Server Error");
        } elseif ($request->header('X-Mismatch-Finder-Not-Found') == $request->path()) {
            abort(404);
```

itamargiv (commented):
Looks good in general, one small non-blocking suggestion, let me know what you prefer

itamargiv (approved):
+1 Thank you

PR #224, opened 2021-11-22
Setup auto generated spec with L5-swagger package

guergana:
Please review by commit. Otherwise it's going to be confusing. :)

You can test by checking the [/api-docs](http://localhost:3000/api-docs/)

Bug: [T290066](https://phabricator.wikimedia.org/T290066)


guergana:
Closing in favor of keep [the manually crafted yml file](https://github.com/wmde/wikidata-mismatch-finder/blob/main/public/api-docs/mismatchfinder-api-spec.yml). We will make sure the spec gets update by writing some contract tests that check that the yml file includes the endpoints defined in the application.

PR #229, opened 2021-11-26
Display generic error message on failed requests

itamargiv:
This change adds a flag to display error messages derived from failed requests (status code >400).

Bug: [T295990](https://phabricator.wikimedia.org/T295990)


Silvan-WMDE (approved):
paired, tested locally - works. :100:

PR #238, opened 2021-12-01
Fix CI Issues

itamargiv:
This change pins the PHP version of each of our jobs to PHP 7.3 (same version as toolforge). Additionally, some redundant action dependencies are removed for consistency.

Bug: [T296846](https://phabricator.wikimedia.org/T296846)


guergana (approved):
Thanks for doing this!

PR #241, opened 2021-12-06
Fix outdated test for (not) submitting reviews

Silvan-WMDE:
This change adjusts a vue unit test to the switch from inertia to axios
that was done in PR #159 (commit ba5b704198ab). It must have been missed
back then.

Bug: T293730

### DIFF
--- a/tests/Vue/Pages/Results.spec.js
+++ b/tests/Vue/Pages/Results.spec.js
@@ -238,15 +238,16 @@ describe('Results.vue', () => {
     });
 
     it('Does not send a put request without any decisions', () => {
+        // clear mock object
+        axios.put = jest.fn();
+
         const item_id = 'Q321';
         const decisions = { [item_id]:{1:{id:1, item_id ,review_status: ReviewDecision.Wikidata}}};
-        const inertiaPut = jest.fn();
         const wrapper = mountWithMocks({
-            mocks: { $inertia: { put: inertiaPut } },
             data: { decisions }
         });
         wrapper.vm.send( 'Q42' );
-        expect( inertiaPut ).not.toHaveBeenCalled();
+        expect( axios.put ).not.toHaveBeenCalled();
 
         //the decisions object will be untouched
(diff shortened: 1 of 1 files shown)


tests/Vue/Pages/Results.spec.js
```diff
@@ -238,15 +238,16 @@ describe('Results.vue', () => {
...
+
         const item_id = 'Q321';
         const decisions = { [item_id]:{1:{id:1, item_id ,review_status: ReviewDecision.Wikidata}}};
-        const inertiaPut = jest.fn();
         const wrapper = mountWithMocks({
-            mocks: { $inertia: { put: inertiaPut } },
             data: { decisions }
         });
         wrapper.vm.send( 'Q42' );
-        expect( inertiaPut ).not.toHaveBeenCalled();
```
itamargiv, line 249: That's an interesting case, if you consider the reason it was missed, is because the assertion is a negative one, I wonder if a test like this is really useful at the end of the day.

Silvan-WMDE: I understand your doubts, but I would like to leave it in there, because this way the test covers both the user perspective (decisions will be untouched) and the system's perspective (don't bother me with API requests if you have nothing to tell). Currently these two calls directly follow one another in `Results.vue`s try-catch block, but I think for future refactoring of the `send` method, the test will make sure that both will happen.

The fact that we forgot to switch the test from inertia to axios when the implementation changed, is some kind of bad luck, I guess. To prevent that, we'd probably have to change the `send` method from `void` to returning something (the `put` response? or the HTTP status?) and then test that it matches some kind of `null` expectation in case the guard clause was triggered. Do you think that would be better?

Silvan-WMDE: PS: I know it is not a sanity proof, but [here](https://stackoverflow.com/q/24282390) someone asks on SO for the specific `not.toHaveBeenCalled()` functionality and of all the people who liked the answer or the question not a single one left a comment that this shouldn't be done. :-)

itamargiv (approved):
LGTM, thank you!

PR #242, opened 2021-12-06
Add fix to not submit reverted reviews

Silvan-WMDE:
Bug: [T297015](https://phabricator.wikimedia.org/T297015)

### DIFF
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -313,7 +313,7 @@
                 this.lastSubmitted = item;
             },
             hasChanged(entityId: string) {
-                for (var decisionId in this.decisions[entityId]) {
+                for (const decisionId in this.decisions[entityId]) {
                     const decision = this.decisions[entityId][decisionId];
                     if(decision.review_status !== decision.previous_status) {
                         return true;
@@ -323,7 +323,7 @@
                 return false;
             },
             storePreviousDecisions(item: string) {
-                for (var decisionId in this.decisions[item]) {
+                for (const decisionId in this.decisions[item]) {
                     const decision = this.decisions[item][decisionId];
                     decision.previous_status = decision.review_status;
                 }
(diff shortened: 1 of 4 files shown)


resources/js/Pages/Results.vue
```diff
@@ -149,7 +150,8 @@
     interface MismatchDecision {
         id: number,
         item_id: string,
-        review_status: ReviewDecision
+        review_status: ReviewDecision,
+        old_status: ReviewDecision
```
guergana, line 154: Works great, thanks!!. A nitpick: `previous_status` is a more appropriate name than `old_status`.

resources/js/Pages/Results.vue
```diff
@@ -309,6 +312,22 @@
...
+                for (var decisionId in this.decisions[entityId]) {
+                    const decision = this.decisions[entityId][decisionId];
+                    if(decision.review_status !== decision.previous_status) {
+                        return true;
+                    }
+                }
+
+                return false;
+            },
+            storeOldDecisions(item: string) {
```
guergana, line 325: here too storePreviousDecisions for consistency. Sorry, just noticed this one :sweat_smile:

resources/js/Pages/Results.vue
```diff
@@ -309,6 +312,22 @@
...
+                    const decision = this.decisions[entityId][decisionId];
+                    if(decision.review_status !== decision.previous_status) {
+                        return true;
+                    }
+                }
+
+                return false;
+            },
+            storePreviousDecisions(item: string) {
+                for (var decisionId in this.decisions[item]) {
```
guergana, line 326: Sorry, I didn't catch this one before... It's better to use a `const` here. In any case `let` or `const` to be clear about scoping.

resources/js/Pages/Results.vue
```diff
@@ -309,6 +312,22 @@
             showSubmitConfirmation( item: string ) {
                 this.lastSubmitted = item;
             },
+            hasChanged(entityId: string) {
+                for (var decisionId in this.decisions[entityId]) {
```
guergana, line 316: also here `var` should be `const`

guergana (changes requested):

guergana (approved):
LGTM!!!

PR #243, opened 2021-12-06
Results page - Loading state after submitting the changes in the dropdowns

guergana:
This PR adds a loading overlay and progressbar after submitting result changes taking the styles used in Wikit's Dialog overlay and Wikit's TextArea progressbar.

Bug: [T293871](https://phabricator.wikimedia.org/T293871)


Silvan-WMDE (approved):
Works and looks great. :100:

Now that I see the solution, it really makes a lot of sense to put it all into a separate component. To me, it feels like everything is exactly where I'd expect it to be. :+1:

PR #252, opened 2021-12-09
Replace local Dialog with WiKit component

itamargiv:
This change replaces the local dialog component with the WiKit one, initially released in WiKit v2.1.0-alpha.9.

Bug: [T295877](https://phabricator.wikimedia.org/T295877)


Silvan-WMDE (approved):
After so many components that we've added to Wikit, I still love these PRs. :-) :+1:

PR #259, opened 2021-12-13
Add user id to results page response

itamargiv:
This change includes a small fix to a bug detected while reviewing another PR. The user ID was not actually included in the response to the results page, which causes the confirmation dialogs checkbox value to be saved in local storage under a key that does not differentiate between users. This meant that two separate users on the same machine were actually sharing the same setting (if the decided not to display the confirmation message again).

Bug: [T297608](https://phabricator.wikimedia.org/T297608)


Silvan-WMDE (approved):
Nice catch, and it works: now the local storage key contains the user id, rather than `undefined`. :100:

PR #262, opened 2021-12-14
Simplify logic, Fix Browser Tests

itamargiv:
This change reverts the logic in the send method to its former procedure, but also adds logic to disable the submitting state, and reduces the flash wait to 250ms which provides a more subtle transition without making the user wait at least 1 second for fast requests.

In addition, an unnecessary browser test is removed, and pause times have been added to allow the confirmation dialog to finish animating before assertions.

### DIFF
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -179,6 +179,8 @@
         submitting: boolean
     }
 
+    const delay = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
+
     export default defineComponent({
         components: {
             Head,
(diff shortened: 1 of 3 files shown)


resources/js/Pages/Results.vue
```diff
@@ -289,42 +292,37 @@
...
+                    // adding this delay because when the response from the request happens
+                    // too fast the overlay and progressbar flash
+                    setTimeout(() => {
+                        this.submitting = false;
+                        document.body.classList.remove('noscroll');
 
+                        if(!this.disableConfirmation){
+                            confirmationDialog.show();
+                        }
+                    }, 250);
```
itamargiv, line 314: This is the main reason why I think the overlay should get it's own component. Instead of repeating this wait each time, we can encapsulate this within the component itself with clever usage of props and events.

guergana: 250 is also fine. I chose 1s as an arbitrary number. Why not put this in a variable if we are using it in more than one place, though?

resources/js/Pages/Results.vue
```diff
@@ -281,6 +279,11 @@
 
                 this.clearSubmitConfirmation();
 
+                this.submitting = true;
```
guergana, line 282: :P what is exactly the difference between having this here and a few lines below?

resources/js/Pages/Results.vue
itamargiv: Because this is all going to go to a new component I'm working on now.

resources/js/Pages/Results.vue
itamargiv: no difference, it's the other lines that shouldn't have been before the axios call.

resources/js/Pages/Results.vue
guergana: ah, alrighty. :)

resources/js/Pages/Results.vue
```diff
@@ -179,7 +179,7 @@
         submitting: boolean
     }
 
-    const SUBMITTING_DELAY_TIME = 1000;
+    const delay = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
```
guergana, line 182: Fancy

itamargiv: ha, well it's for to make sure that the send function doesn't finish resolving until the loading animation fully completes at least 250ms (before the tests were failing because the setTimeout was finished only after the send promise resolved).

guergana (approved):

PR #265, opened 2021-12-14
Extract loading overlay to a component

itamargiv:
This change takes the loading overlay created by @guergana and extracts it to its own component with `show` and `hide` methods, as suggested by @Silvan-WMDE in a call last week.

### DIFF
--- a/resources/js/Components/LoadingOverlay.vue
+++ b/resources/js/Components/LoadingOverlay.vue
@@ -23,7 +23,7 @@ export default Vue.extend({
     data() {
         return {
             shown: this.visible,
-            document: {
+            cachedStyles: {
                 overflow: 'auto'
             }
         };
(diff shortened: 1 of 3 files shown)


tests/Vue/Components/LoadingOverlay.spec.js
```diff
@@ -0,0 +1,44 @@
...
+    });
+
+    it('hides when calling hide method', () => {
+        const wrapper = mount(LoadingOverlay);
+
+        // Hide returns a promise to ensure the animation completes at least after a minimal delay
+        return wrapper.vm.hide()
+            .then(() => wrapper.vm.$nextTick())
+            .then(() => {
+                expect(wrapper.find('.loading-indicator').exists()).toBe(false);
```
guergana, line 41: nice test as well

resources/js/Components/LoadingOverlay.vue
```diff
@@ -0,0 +1,123 @@
...
+    methods: {
+        show(): void {
+            if (this.shown) {
+                return;
+            }
+
+            const bodyStyles = window.getComputedStyle(document.body);
+            this.shown = true;
+
+            this.document.overflow = bodyStyles.overflow;
```
guergana, line 40: It works great. Thanks. I am having trouble understanding why this part was implemented like this

itamargiv: I want to avoid adding an extra css class to an html element that is not is the scope of this template, so We change the overflow programmatically. To do so, we cache this value so that we don't accidentally override any styles that were applied to the document in html, css or other scripts. This is so that the component itself is not too closely coupled to the current implementation (which implicitly assumes that we actually can override the body's overflow with just a class).

guergana: Yes, it's a good solution to go back to whatever styles where applied before.

guergana (approved):
Thank you. 😊

PR #267, opened 2021-12-15
Request entity labels by batches of 50

itamargiv:
This change ensures that any requests made to `wbformatentities` does not surpass the 50 entity ID limit stipulated by the API. In addition, a generic exception was added to the `WikibaseAPIClient` class so that it would throw on any `GET` request that results in an error.

Bug: [T297774](https://phabricator.wikimedia.org/T297774)


Silvan-WMDE (approved):
Looks great. :100:

Tested locally and confirmed that entities are sent to wikidata API in batches.

PR #268, opened 2021-12-15
Validate maximal amount of Item IDs on the client side

itamargiv:
This change adds client side validation logic to warn users if they input more than 600 IDs in the item ID form. Additionally, the client side validation logic has been slightly restructured to allow for easier expansion or modification of the form's validation rules.

Bug: [T295019](https://phabricator.wikimedia.org/T295019)

### DIFF
--- a/tests/Vue/Pages/Home.spec.js
+++ b/tests/Vue/Pages/Home.spec.js
@@ -67,7 +67,7 @@ describe('Home.vue', () => {
     it('validates that items in textarea input dont exceed the maximum number of ids', async () => {
         const store = new Vuex.Store();
 
-        const itemsInput = rangeIds(1, MAX_NUM_IDS + 2).join('\n');
+        const itemsInput = Array(MAX_NUM_IDS + 2).fill('Q21').join('\n');
 
         const wrapper = mount(Home, {
             mocks,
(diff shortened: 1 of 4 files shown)


guergana (commented):
Hiii. Maybe we want to add a test for this???

guergana (commented):
Great work otherwise. I have tested it manually and it works like a charm :)

tests/Vue/Pages/Home.spec.js
```diff
@@ -64,4 +64,34 @@ describe('Home.vue', () => {
         expect(dialog.isVisible()).toBe(true);
     });
 
+    it('validates that items in textarea input dont exceed the maximum number of ids', async () => {
+        const store = new Vuex.Store();
+
+        const itemsInput = rangeIds(1, MAX_NUM_IDS + 2).join('\n');
```
guergana, line 70: started from 1 because testing locally for Q0 has caused errors for me in the past.

itamargiv: Nice! The test looks great, thank you! I think we don't actually have to increment the QIDs, since the sanitization function doesn't deduplicates the IDs in the list, so instead of adding a function we can write something like:

```suggestion
        const itemsInput = Array(MAX_NUM_IDS + 2).fill('Q21');
```

guergana (approved):
looks great! And tested ;)

PR #269, opened 2021-12-15
Update dependabot interval

itamargiv:
This change updates Dependabot's interval to avoid spamming the PR list.


guergana (approved):
Thanks!

PR #271, opened 2021-12-21
Handle server side validation errors in client

Silvan-WMDE:
This change adds a computed property to the Home page, which will trigger an error message box in case of serverside validation errors of requested item IDs. Additionally it adds one unit tests to check that the error message box is also shown on the Results page in case of a failed axios PUT request to submit review decisions, as well as one to check that the box disappears again after a successful submit request.

Bug: [T295541](https://phabricator.wikimedia.org/T295541)

### DIFF
--- a/tests/Vue/Pages/Results.spec.js
+++ b/tests/Vue/Pages/Results.spec.js
@@ -225,7 +225,7 @@ describe('Results.vue', () => {
 
     });
 
-    it('Handles errors on axios put requests gracefully', async () => {
+    it('Handles errors on axios PUT requests gracefully', async () => {
         // mock axios error response
         axios.put = jest.fn().mockRejectedValue('Error');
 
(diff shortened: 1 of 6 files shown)


tests/Vue/Pages/Results.spec.js
```diff
@@ -241,6 +241,32 @@ describe('Results.vue', () => {
...
+    it('Shows error message on failed axios PUT request', async () => {
+        const wrapper = mountWithMocks({
+            data: { 'requestError' : true }
+        });
+
+        const errorMessage = wrapper.find('#error-section .wikit-Message--error.wikit');
+        expect(errorMessage.isVisible()).toBe(true);
+    });
+
+    it('Clears error message on successful axios PUT request', async () => {
```
guergana, line 253: Great! thanks

guergana (approved):
Works great. Thanks.

PR #272, opened 2021-12-30
Metrics/requests

guergana:
TODO: ~~find out why stats are being sent from the Results controller for `mismatch_request` but same code is not sending anything from MismatchController:get when accessed through the web endpoint. Do I need to make some code injection somewhere?~~ please double check. I think this is working now.  

The stats can be seen in [Graphite](https://graphite.wikimedia.org/?showTarget=Wikidata.mismatch-finder.mismatch_review.rate&width=586&height=308&target=Wikidata.mismatch-finder.mismatch_review.lower&target=Wikidata.mismatch-finder.mismatch_review.mean&target=Wikidata.mismatch-finder.mismatch_review.count) and [Grafana](https://grafana-rw.wikimedia.org/d/knCj7WA7z/wikidata-mismatch-finder?orgId=1)

MismatchController is sending the stats from the index and update methods in the api. 

UPDATE: i am not aware at the moment on the other metrics we need but maybe it would make sense to send also the operation to the `sendStats` function? In this case `increment` but the function can also be modified if needed in the future and not necessarily in this PR. This PR assumes we would only want to send increments to graphite

Bug: [T295758](https://phabricator.wikimedia.org/T295758)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -58,3 +58,6 @@ PUSHER_APP_CLUSTER=mt1
 
 MIX_PUSHER_APP_KEY="${PUSHER_APP_KEY}"
 MIX_PUSHER_APP_CLUSTER="${PUSHER_APP_CLUSTER}"
+
+STATSD_HOST='https://www.wikidata.org/beacon/statsv'
+STATSD_NAMESPACE='Wikidata.mismatch-finder'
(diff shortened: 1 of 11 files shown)


Silvan-WMDE (approved):
LGTM - great work, thanks! :100:

PR #280, opened 2022-01-03
Use trait for statsd tracking

Silvan-WMDE:
This is just an idea, let's discuss...

### DIFF
--- a/app/Http/Controllers/ResultsController.php
+++ b/app/Http/Controllers/ResultsController.php
@@ -9,7 +9,6 @@ use Illuminate\Support\Facades\Auth;
 use App\Models\Mismatch;
 use App\Services\StatsdAPIClient;
 use Illuminate\Support\Facades\App;
-use Illuminate\Support\Facades\Log;
 use Inertia\Response;
 use Illuminate\Support\LazyCollection;
 
@@ -58,8 +57,6 @@ class ResultsController extends Controller
             $mismatches->isNotEmpty() ? [ 'results' => $mismatches->groupBy('item_id') ] : []
         );
 
-        //collect metric
-        Log::info("statsd 'mismatch_request' (via web)");
         $this->trackRequestStats();
 
         return inertia('Results', $props);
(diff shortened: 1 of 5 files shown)


app/Services/StatsdAPIClient.php
```diff
@@ -17,6 +17,11 @@ class StatsdAPIClient
     public function __construct(string $baseUrl, string $namespace)
     {
         $this->baseUrl = $baseUrl;
+
+        if (!app()->environment('production')) {
```
guergana, line 21: What does this do?

app/Http/Controllers/Traits/StatsTracker.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace App\Http\Controllers\Traits;
+
+trait StatsTracker
+{
+    public function trackRequestStats()
```
guergana, line 7: this is great!! Thanks.

app/Http/Controllers/ResultsController.php
```diff
@@ -55,6 +58,10 @@ public function index(MismatchGetRequest $request, WikibaseAPIClient $wikidata):
             $mismatches->isNotEmpty() ? [ 'results' => $mismatches->groupBy('item_id') ] : []
         );
 
+        //collect metric
+        Log::info("statsd 'mismatch_request' (via web)");
```
guergana, line 62: If we want to merge this maybe we should remove the log? I can imagine the log getting cluttered really quickly once we being to use this in prod.

guergana (changes requested):
Great!!. I dont know if this is a best-practice, though, as we discussed. 
I would remove the logs for now.

app/Services/StatsdAPIClient.php
Silvan-WMDE: it checks if the app is running in production, and if not it uses the "mismatch-finder-test" namespace, so that our CI tests as well as the dev and staging environments don't interfere with the production stats. Check the [Grafana board](https://grafana-rw.wikimedia.org/d/knCj7WA7z/wikidata-mismatch-finder?orgId=1&from=now-7d&to=now) for my newly introduced "(testing)" metrics in light purple.

app/Http/Controllers/ResultsController.php
Silvan-WMDE: yes, definitely. I introduced the logs for myself so that I can verify what's going on and then let them in there, so you can use them to understand what I changed. But I agree, we should definitely remove them again.

app/Services/StatsdAPIClient.php
guergana: Awesome!!!! Great addition.

PR #281, opened 2022-01-10
Add Button to download review statistics for all imports

guergana:
This PR adds a button to the Import Status Section in the store that upon click lets the user download a CSV file with statistics of the uploades imports.

TODO: write tests, feature and browser.

Bug: [T295759](https://phabricator.wikimedia.org/T295759)

### DIFF
--- a/app/Http/Controllers/CsvController.php
+++ b/app/Http/Controllers/CsvController.php
@@ -56,17 +56,27 @@ class CsvController extends Controller
         foreach ($imports as $each_import) {
             $mismatches = Mismatch::where('import_id', $each_import->id)->get();
             $mismatches_count = $mismatches->count();
+            $review_statuses = array( 'wikidata', 'external', 'both','none', 'pending' );
+            $percent_completed = 0;
 
+            foreach ($review_statuses as $review_status) {
+                ${"mismatches_" . $review_status} = $mismatches->where('review_status', $review_status)->count();
+            }
+
+            if ($mismatches_count > 0) {
+                $percent_completed = (($mismatches_count - $mismatches_pending) / $mismatches_count) * 100;
+            }
+    
             fputcsv($handle, [
                 $each_import->id,
                 $each_import->status,
                 $mismatches_count,
-                100,
-                100,
-                100,
-                100,
-                100,
-                100
+                $mismatches_wikidata,
+                $mismatches_external,
+                $mismatches_both,
+                $mismatches_none,
+                $mismatches_pending,
+                $percent_completed
             ]);
         }
         fclose($handle);
--- a/app/Http/Controllers/CsvController.php
+++ b/app/Http/Controllers/CsvController.php
@@ -3,26 +3,21 @@
 namespace App\Http\Controllers;
 
 use Illuminate\Support\Facades\Response;
-use App\Models\ImportMeta;
-use App\Services\CSVImportStatsFileGenerator;
+use App\Services\ImportReport;
 use Carbon\Carbon;
 
 class CsvController extends Controller
 {
-    
-    public function download_csv()
-    {
 
-        $imports = ImportMeta::get();
+    public function download_csv(ImportReport $report)
+    {
         $currentTime = Carbon::now()->format('Y-m-d');
 
         $headers = array(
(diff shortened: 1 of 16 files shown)


app/Http/Controllers/CsvController.php
```diff
@@ -0,0 +1,75 @@
...
+        ]);
+
+        foreach ($imports as $each_import) {
+            $mismatches = Mismatch::where('import_id', $each_import->id)->get();
+            $mismatches_count = $mismatches->count();
+            $review_statuses = array( 'wikidata', 'external', 'both','none', 'pending' );
+            $percent_completed = 0;
+
+            foreach ($review_statuses as $review_status) {
+                ${"mismatches_" . $review_status} = $mismatches->where('review_status', $review_status)->count();
```
itamargiv, line 52: ooh, clever! I like it :+1:

app/Http/Controllers/CsvController.php
```diff
@@ -0,0 +1,75 @@
...
+{
+    
+    public function get_csv()
+    {
+
+        $imports = ImportMeta::get();
+        $currentTime = Carbon::now()->format('Y-m-d');
+
+        $headers = array(
+            'Content-Type' => 'application/vnd.ms-excel; charset=utf-8'
```
itamargiv, line 21: I think we should  use `text/csv` here to make this as interoperably and standardized as possible. See: https://stackoverflow.com/a/7076079

resources/views/components/layout.blade.php
```diff
@@ -7,13 +7,13 @@
     <title>{{__('store-layout.mismatch-finder')}}</title>
     <link rel="stylesheet" href="{{ asset('css/app.css') }}">
 </head>
-<body>
+<body class="{{ Route::currentRouteName() }}">
```
itamargiv, line 10: Smart!

itamargiv (changes requested):
Looks fantastic, thank you! Generally speaking, I think we should move the csv report generation logic to its own class, so that we don't mix business logic into the controller that created the download response. Also, I think we should test this (both with feature and browser tests), to make sure that a report is generated and that users are able to download it.

app/Http/Controllers/CsvController.php
guergana: yup i agree.

tests/Feature/ImportReportTest.php
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * @return void
+     */
+    public function test_generate_csv_writes_to_path(): void
+    {
+        $user = User::factory()->uploader()->create();
+        $import = ImportMeta::factory()->for($user)->create(
+            ['status' => 'completed']
+        );
+        Mismatch::factory()->for($import)->create();
```
Silvan-WMDE, line 26: I think at this point I would have added at least two mismatches, one pending and one reviewed, if only to show that the "% completed" calculation is done correctly.

Silvan-WMDE (dismissed):
Looks good to me. I agree browser tests are expendable and I like the feature tests. Just added a minor suggestion to one of them, no blocker. Will happily re-approve if you decide to pick it up.

tests/Feature/ImportReportTest.php
itamargiv: done, I even added three so that it would b 33.33% vs 66.67% :D

tests/Feature/ImportReportTest.php
```diff
@@ -0,0 +1,82 @@
...
+                $import->id, // Import ID
+                $import->status, // Import status
+                $total, // Mismatch count
+                $mismatches->where('review_status', 'wikidata')->count(), // Error on wikidata count
+                $mismatches->where('review_status', 'external')->count(), // Error on external count
+                $mismatches->where('review_status', 'both')->count(), // Error on both count
+                $mismatches->where('review_status', 'none')->count(), // Error on none count
+                $pending, // Pending count
+                100 - ($pending / $total * 100), // % completed
+            ]
```
Silvan-WMDE, line 57: nice, I like!

Silvan-WMDE (approved):
Looks good to me - thanks for the work! :+1:

PR #283, opened 2022-01-12
Track open vs reviewed mismatches via statsv

Silvan-WMDE:
This change adds a scheduled tracking task that runs every minute. It sends the
total number of mismatches and the number of reviewed mismatches to
statsv for reporting purposes.

Bug: [T295760](https://phabricator.wikimedia.org/T295760)

See "Total vs Reviewed" panel on [Grafana Dashboard](https://grafana-rw.wikimedia.org/d/knCj7WA7z/wikidata-mismatch-finder)

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -2,6 +2,7 @@
 
 namespace App\Console;
 
+use App\Console\Tasks\TrackMismatchRatio;
 use Illuminate\Console\Scheduling\Schedule;
 use Illuminate\Foundation\Console\Kernel as ConsoleKernel;
 
@@ -24,7 +25,9 @@ class Kernel extends ConsoleKernel
      */
     protected function schedule(Schedule $schedule)
     {
-        // $schedule->command('inspire')->hourly();
+        $schedule->call(new TrackMismatchRatio())
+            ->description("Tracking task that runs every minute")
+            ->everyMinute();
     }
 
     /**
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -26,8 +26,8 @@ class Kernel extends ConsoleKernel
     protected function schedule(Schedule $schedule)
     {
         $schedule->call(new TrackMismatchRatio())
-            ->description("Tracking task that runs every minute")
-            ->everyMinute();
+            ->description("Task to track the 'pending vs reviewed' ratio once per day")
+            ->daily();
     }
 
     /**
--- a/app/Http/Controllers/Traits/StatsTracker.php
+++ b/app/Http/Controllers/Traits/StatsTracker.php
@@ -18,4 +18,18 @@ trait StatsTracker
     {
         $this->statsd->sendStats('import_mismatch_file');
     }
+
+    /**
+     * Track the current 'total vs reviewed' mismatch ratio via statsv
+     *
+     * @param int $total    Total number of unexpired mismatches
+     * @param int $reviewed Total number of unexpired mismatches that are reviewed
+     *
+     * @return void
+     */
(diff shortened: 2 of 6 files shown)


app/Services/StatsdAPIClient.php
```diff
@@ -25,9 +25,9 @@ public function __construct(string $baseUrl, string $namespace)
         $this->namespace = $namespace;
     }
 
-    public function sendStats(string $metric): Response
+    public function sendStats(string $metric, int $value = 1): Response
     {
-        $response = Http::post($this->baseUrl. '?' .$this->namespace . '.' . $metric . '=1c');
+        $response = Http::post($this->baseUrl. '?' .$this->namespace . '.' . $metric . '=' . $value . 'c');
```
guergana, line 30: Nice!

app/Http/Controllers/Traits/StatsTracker.php
```diff
@@ -4,18 +4,47 @@
 
 trait StatsTracker
 {
+    /**
```
guergana, line 7: This is good..For the docs.

app/Services/StatsdAPIClient.php
```diff
@@ -25,9 +25,9 @@ public function __construct(string $baseUrl, string $namespace)
         $this->namespace = $namespace;
     }
 
-    public function sendStats(string $metric): Response
+    public function sendStats(string $metric, int $value = 1): Response
```
guergana, line 28: Good stuff.

app/Console/Kernel.php
```diff
@@ -24,7 +25,9 @@ class Kernel extends ConsoleKernel
      */
     protected function schedule(Schedule $schedule)
     {
-        // $schedule->command('inspire')->hourly();
+        $schedule->call(new TrackMismatchRatio())
+            ->description("Tracking task that runs every minute")
+            ->everyMinute();
```
guergana, line 30: why is it every minute? isn't that too frequent? do we need that level of granularity? i may have missed or forgotten anything related to the frecuency of the scheduler.

Silvan-WMDE: Good question, let me find out. Most probably it isn't even needed every minute.

app/Console/Kernel.php
```diff
@@ -24,7 +25,9 @@ class Kernel extends ConsoleKernel
      */
     protected function schedule(Schedule $schedule)
     {
-        // $schedule->command('inspire')->hourly();
+        $schedule->call(new TrackMismatchRatio())
```
itamargiv, line 28: This is a really great feature!

containers/production-schedule.yaml
```diff
@@ -0,0 +1,31 @@
+---
+# Run the mismatch finder scheduler on kubernetes
+apiVersion: apps/v1
+kind: Deployment
```
itamargiv, line 4: Is there a particular reason de decided to go with a deployment rather than a [`kind: CronJob`](https://wikitech.wikimedia.org/wiki/Help:Toolforge/Kubernetes#Kubernetes_cronjobs) pod here? I'm actually fine with both, but I wonder if there is a tradeoff between the two approaches.

itamargiv (commented):
This looks fantastic, thank you! I only have one question, but otherwise, it's good to go!

containers/production-schedule.yaml
Silvan-WMDE: no you are right: I didn't pay attention and actually agree a cronjob that runs every minute is better than the continuous `schedule:work` job.

itamargiv (approved):
Fantastic job, thank you!

PR #286, opened 2022-01-20
Add expiry data columns to report

itamargiv:
This change adds import expiration data as a followup to the import overview report.

Bug: [T295759](https://phabricator.wikimedia.org/T295759)


Silvan-WMDE (approved):
Looks good to me - thank you!

PR #287, opened 2022-01-24
Extract vuex implementation for consistency and testability

guergana:
Bug: [T295704](https://phabricator.wikimedia.org/T295704)

### DIFF
--- a/resources/js/store/index.ts
+++ b/resources/js/store/index.ts
@@ -1,5 +1,6 @@
 import Vue from 'vue';
 import Vuex, {Store} from 'vuex';
+import RootState from './RootState';
 import { Inertia } from '@inertiajs/inertia';
 import mutations from './mutations';
 
--- a/resources/js/store/index.ts
+++ b/resources/js/store/index.ts
@@ -4,7 +4,7 @@ import RootState from './RootState';
 import { Inertia } from '@inertiajs/inertia';
 import mutations from './mutations';
 
-export function getInitialState() {
+export function getInitialState(): RootState {
     return {
         loading: false,
         lastSearchedIds: ''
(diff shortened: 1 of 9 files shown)


resources/js/store/index.ts
```diff
@@ -0,0 +1,40 @@
+import Vue from 'vue';
+import Vuex, {Store} from 'vuex';
+import { Inertia } from '@inertiajs/inertia';
+import mutations from './mutations';
+
+export function getInitialState() {
```
itamargiv, line 6: Nice one, we can use RootState in here as well.

resources/js/store/index.ts
```diff
@@ -0,0 +1,40 @@
...
+    return {
+        loading: false,
+        lastSearchedIds: ''
+    }
+}
+
+export function createStore(): Store<{loading: boolean}>{
```
itamargiv, line 13:
```suggestion
export function createStore(): Store<RootState>{
```

itamargiv (commented):
Great job, thank you! One small comment but non blocker

resources/js/store/index.ts
guergana: oh weird. i dont remember adding that. maybe it was automatically added.

resources/js/store/index.ts
guergana: But RootState is just the types. getInitialState returns the initial values. I think they are different things.

itamargiv: Yes, I mean that we should annotate the return type of this function with RootState, that's what it returns after all. No?

guergana: yes! Excuse moi, I misunderstood your comment :)

guergana: Annotated!

itamargiv (approved):
Thank yout, looks great!

resources/js/store/index.ts
itamargiv: np, probably wasn't clear enough. Thx!

PR #305, opened 2022-02-01
Add local job queue running instructions

itamargiv:
This change adds additional instructions on how to run the job queue locally.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -199,6 +199,7 @@ By Default, the jobs will run synchronously to the requests that dispatch them. 
     sail artisan queue:listen --timeout=1200
     ```
 
+    The `--timeout=1200` flag in the command above increases the job timeout to 20 minutes thus ensuring that the queue worker doesn't abort jobs that take longer than a minute.
 ## Linting <a id="linting"></a>
 ### PHP Linting <a id="php-linting"></a>
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -9,11 +9,13 @@
   - [Start the application server](#start-the-application-server)
   - [Stop the application server](#stop-the-application-server)
   - [Destroy all the things](#destroy-all-the-things)
-- [Working with OAuth](#working-with-oauth)
-- [Frontend - Working with CSS and JS](#frontend---working-with-css-and-js)
+- [Working with OAuth](#oauth)
+- [Frontend - Working with CSS and JS](#frontend)
+- [Localization and Internationalization](#localization-and-internationalization)
+- [Job Queues](#job-queues)
 - [Linting](#linting)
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -9,11 +9,13 @@
   - [Start the application server](#start-the-application-server)
   - [Stop the application server](#stop-the-application-server)
   - [Destroy all the things](#destroy-all-the-things)
-- [Working with OAuth](#working-with-oauth)
-- [Frontend - Working with CSS and JS](#frontend---working-with-css-and-js)
+- [Working with OAuth](#oauth)
+- [Frontend - Working with CSS and JS](#frontend)
+- [Localization and Internationalization](#localization-and-internationalization)
```
guergana, line 14: :thinking: Where is this coming from?

docs/README.md
```diff
@@ -183,6 +185,20 @@ To ensure that your client side localization files are valid, run:
...
+As the Mismatch Finder includes jobs to validate and import mismatch CSV files, it is possible to also configure the job queue runner to run locally.
+
+By Default, the jobs will run synchronously to the requests that dispatch them. However, in order to queue and run jobs from the database, simple follow these two steps:
+
+1. In your local `.env` file, change the value of `QUEUE_CONNECTION` to `database` (it should be set to `sync` by default).
+
+1. Start the job queue runner by typing in the following command: 
+
+    ```
+    sail artisan queue:listen --timeout=1200
```
guergana, line 199: Maybe it would be more complete if you indicate briefly why `1200` is a good number to put there?

docs/README.md
itamargiv: from `doctoc`, it seems like we didin't run it in this file for a while

docs/README.md
itamargiv: Good point, thank you!

docs/README.md
guergana: I see... But I guess the link to the `href="localization-and-internationalization" should be added as well in the section?

itamargiv: Ah no, in github it works like that by convention, the reason the other's have links is because (at least at the beginning) some of them were shorter than the auto-generated ones.

docs/README.md
```diff
@@ -183,6 +185,21 @@ To ensure that your client side localization files are valid, run:
...
+    ```
+
+    The `--timeout=1200` flag in the command above increases the job timeout to 20 minutes thus ensuring that the queue worker doesn't abort jobs that take longer than a minute.
```
guergana, line 202: Nice thanks. :)

guergana (approved):
Thanks!

PR #306, opened 2022-02-06
containers: Set `concurrencyPolicy: Forbid`

supertassu:
This prevents a never-ending job execution from creating thousands of Kubernetes Job objects causing operational issues on the cluster itself (see T301081 for details and cluster-level preventation).

### DIFF
--- a/containers/production-schedule.yaml
+++ b/containers/production-schedule.yaml
@@ -10,6 +10,7 @@ metadata:
     toolforge: tool
 spec:
   schedule: "* * * * *"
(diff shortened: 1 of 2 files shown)


itamargiv (approved):
Thank you @supertassu for this fix! We weren't aware that this field should be set.

itamargiv:
For future reference, see relevant docs: https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/#concurrency-policy

PR #376, opened 2022-07-11
Update Obtaining an API access token instructions

jmkeil:
correct "Obtaining an API access token", Step 1


lydiapintscher:
Thank you!
Looks good to me.
@itamargiv Could you have a look?

itamargiv (approved):
Thank you!

PR #379, opened 2022-07-27
Run npm audit fix

guergana:
Bug: [T313745](https://phabricator.wikimedia.org/T313745)


micgro42 (approved):

micgro42:
I'm assuming that the tests passing means that the app is essentially working. I haven't tried it locally

PR #413, opened 2022-09-07
Update Jest-related npm dependencies to latest ^28

lucaswerkmeister:
But I don’t think we can upgrade to ^29 yet, it doesn’t look like ts-jest has a matching version for that.

Bug: [T315965](https://phabricator.wikimedia.org/T315965)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
                 "vuex": "^3.6.2"
             },
             "devDependencies": {
-                "@types/jest": "^28.1.6",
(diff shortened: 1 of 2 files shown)


guergana:
> But I don’t think we can upgrade to ^29 yet, it doesn’t look like ts-jest has a matching version for that.
> 
> Bug: [T315965](https://phabricator.wikimedia.org/T315965)

Yes I encountered the same issue when trying to update [Wikidata Query Builder](https://github.com/wikimedia/wikidata-query-builder/).

PR #415, opened 2022-09-08
Remove explicit uuid npm dependency

lucaswerkmeister:
We haven’t imported the package since 450a139894 replaced the local Dialog component (which uses uuid v4 for an HTML ID) with the Wikit version.

Bug: [T315965](https://phabricator.wikimedia.org/T315965)


chukarave (approved):
:+1:

PR #431, opened 2022-10-12
Add controller and endpoint for retrieving 15 random mismatches

guergana:
For easy testing go to the route `/random`. I suggest removing the 
```
->whereHas('importMeta', function ($import) {
       $import->where('expires', '>=', now());
})
``` 
in both `RandomizeController` and `ResultsController` and in case you don't have current non-expired items to test with. 

Bug: [T302289](https://phabricator.wikimedia.org/T302289)

### DIFF
--- a/app/Http/Controllers/RandomizeController.php
+++ b/app/Http/Controllers/RandomizeController.php
@@ -4,6 +4,7 @@ namespace App\Http\Controllers;
 
 use App\Models\Mismatch;
 use App\Services\StatsdAPIClient;
+use Illuminate\Support\Facades\Auth;
 
 class RandomizeController extends Controller
 {
@@ -39,6 +40,13 @@ class RandomizeController extends Controller
                 [ResultsController::class, 'index'],
                 ['ids' => $itemIdsToString]
             );
+        } else {
+            $user = Auth::user() ? [
+                'name' => Auth::user()->username,
+                'id' => Auth::user()->mw_userid
+            ] : null;
+            
+            return inertia('Results', [ 'user' => $user ]);
         }
     }
 }
--- a/app/Http/Controllers/RandomizeController.php
+++ b/app/Http/Controllers/RandomizeController.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use App\Models\Mismatch;
+use App\Services\StatsdAPIClient;
+
+class RandomizeController extends Controller
+{
+    use Traits\ReviewMismatch, Traits\StatsTracker;
+
+    /**
+     * Instantiate a new controller instance.
+     *
+     * @return void
+     */
+    public function __construct(StatsdAPIClient $statsd)
+    {
+        $this->middleware('simulateError');
+        $this->statsd = $statsd;
+    }
+
+    public function index()
+    {
+        $itemIds = Mismatch::with('importMeta.user')
+            ->distinct()
+            ->where('review_status', 'pending')
+            ->whereHas('importMeta', function ($import) {
+                $import->where('expires', '>=', now());
+            })
+            ->inRandomOrder()
+            ->limit(15)
+            ->pluck('item_id')->toArray();
+        
+        if (count($itemIds) > 0) {
+            $itemIdsToString = implode('|', $itemIds);
+
+            return redirect()->action(
+                [ResultsController::class, 'index'],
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/RandomizeController.php
```diff
@@ -0,0 +1,43 @@
...
+        
+        if (count($itemIds) > 0) {
+            $itemIdsToString = implode('|', $itemIds);
+
+            return redirect()->action(
+                [ResultsController::class, 'index'],
+                ['ids' => $itemIdsToString]
+            );
+        }
+    }
```
lucaswerkmeister, line 42: And what happens otherwise? What does Laravel do if we don’t return anything?

app/Http/Controllers/RandomizeController.php
```diff
@@ -0,0 +1,43 @@
...
+    public function index()
+    {
+        $itemIds = Mismatch::with('importMeta.user')
+            ->where('review_status', 'pending')
+            ->whereHas('importMeta', function ($import) {
+                $import->where('expires', '>=', now());
+            })
+            ->inRandomOrder()
+            ->limit(15)
+            ->pluck('item_id')->toArray();
```
lucaswerkmeister, line 32: Is this guaranteed to return 15 different items, or could the same item also be selected several times?

lucaswerkmeister:
> With this proposed implementation we are making now two database queries, one to get the mismatches and then we are again making a request when redirected to the results controller. Would it make sense to try to optimize it?

I don’t think it makes sense to optimize the number of database queries.  We should optimize the queries themselves, so that they’re cheap enough to run however many times we need.

app/Http/Controllers/RandomizeController.php
guergana: I did a manual test and it returns always different items

hasanakg: > I did a manual test and it returns always different items. I would need to test when there are less than 15 items to see what happens

For less than 15 items, it would select the same items with different order imo.

guergana:
> > With this proposed implementation we are making now two database queries, one to get the mismatches and then we are again making a request when redirected to the results controller. Would it make sense to try to optimize it?
> 
> I don’t think it makes sense to optimize the number of database queries.  We should optimize the queries themselves, so that they’re cheap enough to run however many times we need.

Since this means touching the tables, it would make sense to create a separate ticket for that.

app/Http/Controllers/RandomizeController.php
guergana: This is still to be implemented. I will put that in a separate PR since it includes creating a view to display the message saying that there are no matching items and it would make this PR too big. If you think it should go in this one i could it here.

app/Http/Controllers/RandomizeController.php
guergana: For less than 15 items it should display the available ones, it's outlined in the ticket.

app/Http/Controllers/RandomizeController.php
guergana: > And what happens otherwise? What does Laravel do if we don’t return anything?

I have added this now in a new commit.

lucaswerkmeister:
Hm, I’m not sure I agree with the latest changes (making `/results`’ IDs optional) – I feel like this should be handled on the `/random` side, and `/results` with empty IDs should be different?

app/Http/Controllers/RandomizeController.php
lucaswerkmeister: I tried it out with a CSV file based on the example mismatches, and got a URL like this:

http://localhost:1337/results?ids=Q184746%7CQ9391%7CQ9391%7CQ184746%7CQ9391%7CQ184746%7CQ184746%7CQ184746%7CQ184746%7CQ9391%7CQ184746%7CQ184746%7CQ9391%7CQ9391%7CQ184746

The same item IDs are being repeated over and over again, because the query selects the item ID of 15 mismatches, not 15 different item IDs.

app/Http/Controllers/RandomizeController.php
lucaswerkmeister: It looks like Laravel just returns an empty page in that case, which seems fine for now. (We decided in Mattermost to leave the “no results” part to a second pull request.)

lucaswerkmeister (approved):

PR #432, opened 2022-10-13
Add case for random endpoint when no available unreviewed items

guergana:
Bug: [T302289](https://phabricator.wikimedia.org/T302289)

### DIFF
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -216,7 +216,6 @@
             },
         },
         mounted(){
-            console.log('props', this.$props)
             if(!this.$store.state.lastSearchedIds) {
                 this.saveSearchedIds( this.item_ids.join('\n') );
             }
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -61,6 +61,9 @@
                     </wikit-link>
                 </span>
             </Message>
+            <Message type="notice" v-if="item_ids.length === 0">
+                <span>{{ $i18n('no-mismatches-available-for-review') }}</span>
+            </Message>
             <Message type="warning" v-if="!user">
                 <span v-i18n-html:log-in-message="['/auth/login']"></span>
             </Message>
@@ -213,6 +216,7 @@
             },
         },
         mounted(){
+            console.log('props', this.$props)
             if(!this.$store.state.lastSearchedIds) {
                 this.saveSearchedIds( this.item_ids.join('\n') );
             }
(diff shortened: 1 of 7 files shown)


lucaswerkmeister (commented):
I tried this out locally, and it doesn’t seem to quite work yet – when I go to `/random` when there are no mismatches, I see an empty results page, but the “no mismatches available” message doesn’t show up. So far I can’t figure out why.

I’m also trying to understand how this works (not being familiar with Laravel yet)… the call to `inertia('Results', [ 'user' => $user ])` in `RandomizeController` directly renders `Results.vue`, right? Or does it go through `ResultsController`? (My thinking is that, if it bypasses `ResultsController`, then we shouldn’t need to change the code in there. But maybe I’m misunderstanding something.)

guergana:
> I tried this out locally, and it doesn’t seem to quite work yet – when I go to `/random` when there are no mismatches, I see an empty results page, but the “no mismatches available” message doesn’t show up. So far I can’t figure out why.
> 
> I’m also trying to understand how this works (not being familiar with Laravel yet)… the call to `inertia('Results', [ 'user' => $user ])` in `RandomizeController` directly renders `Results.vue`, right? Or does it go through `ResultsController`? (My thinking is that, if it bypasses `ResultsController`, then we shouldn’t need to change the code in there. But maybe I’m misunderstanding something.)

You are right! Those are leftovers from the rebasing. Sorry about that. Will upload a new version.

lucaswerkmeister:
> I tried this out locally, and it doesn’t seem to quite work yet – when I go to `/random` when there are no mismatches, I see an empty results page, but the “no mismatches available” message doesn’t show up. So far I can’t figure out why.

It turns out I needed to run `vendor/bin/sail npm run watch` (or `dev` instead of `watch`) to update the front-end files.

resources/js/Pages/Results.vue
```diff
@@ -61,6 +61,9 @@
                     </wikit-link>
                 </span>
             </Message>
+            <Message type="notice" v-if="item_ids.length === 0">
```
lucaswerkmeister, line 64: Maybe a comment around this message would make sense? Something like:

```html
<!-- The Results page without item_ids is used by RandomizeController. -->
```

lucaswerkmeister (approved):
Looks pretty good :+1: one optional suggestion.

resources/js/Pages/Results.vue
guergana: ok. i will add it.

guergana: Done.

lucaswerkmeister (approved):
The test looks really nice :+1:

PR #444, opened 2022-10-26
Get property datatypes in ResultsController

lucaswerkmeister:
This is mainly in preparation for parsing and formatting time values, but we can also use it to partially solve T321214, since we now know which properties contain item IDs instead of inferring it from the value. (But note that it doesn’t fully fix T321214, since `addLabels()` in `Results.vue` still adds labels to all values regardless of property data type.)

In the tests, we need to update the factory so that it generates random Wikidata values matching the randomly selected property ID (previously, tests still worked if the two didn’t match). In `ResultsController` / `WebResultsRouteTest`, we go one step further and test three different mismatches with specific data types: `wikibase-item` (should fetch the label), `time` (no special handling yet, but should later be formatted), and `string` (no special handling).

### DIFF
--- a/database/factories/MismatchFactory.php
+++ b/database/factories/MismatchFactory.php
@@ -4,9 +4,23 @@ namespace Database\Factories;
 
 use App\Models\Mismatch;
 use Illuminate\Database\Eloquent\Factories\Factory;
+use InvalidArgumentException;
+use ValueError;
 
 class MismatchFactory extends Factory
 {
+    private const PROPERTIES = [
+        'P580' => 'time',
+        'P582' => 'time',
+        'P585' => 'time',
+        'P50' => 'wikibase-item',
+        'P86' => 'wikibase-item',
+        'P170' => 'wikibase-item',
+        'P225' => 'string',
+        'P742' => 'string',
+        'P2093' => 'string',
+    ];
+
     /**
      * The name of the factory's corresponding model.
      *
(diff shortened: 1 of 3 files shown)


database/factories/MismatchFactory.php
```diff
@@ -4,9 +4,23 @@
 
 use App\Models\Mismatch;
 use Illuminate\Database\Eloquent\Factories\Factory;
+use InvalidArgumentException;
+use ValueError;
 
 class MismatchFactory extends Factory
```
micgro42, line 10: Frankly, somewhere in the path or class-name of this factory it should say that this is about testing and providing fake data. But `database/factories/MismatchFactory.php` and `namespace Database\Factories;` and `class MismatchFactory extends Factory`, all sound very "this is live production code" to me.

Though this PR did not introduce this class, so this is not a blocking comment.

micgro42:
I still have to review the test code, but the production code looks good so far.

database/factories/MismatchFactory.php
```diff
@@ -23,13 +37,44 @@ public function definition()
...
         return [
             'statement_guid' => 'Q' . $this->faker->randomNumber() . '$' . $this->faker->uuid(),
-            'property_id' => 'P' . $this->faker->randomNumber(),
-            'wikidata_value' => $this->getRandomValue(),
+            'property_id' => $this->faker->randomElement(array_keys(self::PROPERTIES)),
+            'wikidata_value' => function (array $attributes) {
+                $propertyId = $attributes['property_id'];
+                if (!array_key_exists($propertyId, self::PROPERTIES)) {
+                    throw new ValueError("Unknown property {$propertyId}, " .
+                        'you have to specify an explicit wikidata_value');
```
micgro42, line 45: This should never happen, because in line 40 we're exactly selecting a key from `self::PROPERTIES`, right?

micgro42 (approved):
Looks good to me

database/factories/MismatchFactory.php
lucaswerkmeister: It might happen if a developer tries to create a mismatch with a custom property ID:

```php
$mismatch = Mismatch::factory()
    ->for($import)
    ->create(['property_id' => 'P12345']);
```

micgro42: Ok, so reading the source code of the `Factory` parent class, it seems the return value of the definition function is used as the basis for a reducer, that I guess somehow play into those `create` calls?

PR #445, opened 2022-10-26
Add sensible default port to .env.example

micgro42:
Instead of letting many new developers run into errors while following the "Quickstart" guide, set a default that has a good chance to work out of the box.

Furthermore, if that APP_PORT env variable is available, it should be part of the example env file in either way, because that file gives the impression of being comprehensive.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,7 +2,8 @@ APP_NAME=Laravel
 APP_ENV=local
 APP_KEY=
 APP_DEBUG=true
-APP_URL=http://localhost
+APP_PORT=1337
+APP_URL="http://localhost:${APP_PORT}"
 ASSET_URL="${APP_URL}"
 
 MEDIAWIKI_OAUTH_CLIENT_ID=


lucaswerkmeister:
I’m curious how the other developers will feel about changing the default port – I think I’m ambivalent about it. (Another approach would be to comment out the port – but I guess that would mean we’d need a commented-out host as well, since in that case the default host shouldn’t contain the port.)

micgro42:
> I’m curious how the other developers will feel about changing the default port – I think I’m ambivalent about it. (Another approach would be to comment out the port – but I guess that would mean we’d need a commented-out host as well, since in that case the default host shouldn’t contain the port.)

What does, from your perspective, speak against using a default port for development that is not 80?

lucaswerkmeister:
Mainly that people might expect http://localhost/ to point to their local mismatch finder, I guess, especially if that’s how it was done so far… (or are most developers already using a custom `APP_PORT`?)

itamargiv:
This seem to have stagnated, should we move this forward?

micgro42:
> This seem to have stagnated, should we move this forward?

This is probably for the people that will work more on this to decide. I'm fine either way.

itamargiv (approved):

PR #446, opened 2022-10-26
Add frontend build instructions to quickstart section of README

guergana:


lucaswerkmeister:
I’m not sure we need to list `npm run watch` here – maybe it’s enough to say `npm run dev`, and then add a kind of “see also” link pointing to the “frontend – working with CSS and JS” section?

micgro42:
Given that the Quickstart includes installation instructions for composer dependencies, it should probably also contain the installation instructions for the node dependencies.

guergana:
> 

You are right. i wasn't aware that they are not already there. will add

micgro42 (commented):
We should probably not be ambiguous in a Quickstart section of the documentation and only give a single command.

I recommend the `npm run dev` one, as that adds the least overhead (both mental and in terms of system load).

For the day-to-day development instructions, one can have a look at the rest of the README.md

micgro42:
> We should probably not be ambiguous in a Quickstart section of the documentation and only give a single command.
> 
> I recommend the `npm run dev` one, as that adds the least overhead (both mental and in terms of system load).
> 
> For the day-to-day development instructions, one can have a look at the rest of the README.md

I didn't see Lucas already wrote something to that effect, sorry

composer.lock
```diff
@@ -124,28 +124,28 @@
         },
         {
             "name": "doctrine/inflector",
-            "version": "2.0.4",
```
micgro42, line 127: This lock file was probably added by accident to the commit and should be removed. Updating dependencies should probably happen in a dedicated commit and PR.

guergana: oh, yes, weird. i dont recall making any composer changes

guergana: it's fixed :)

micgro42 (approved):

PR #448, opened 2022-10-26
Switch to Node 18 (LTS)

micgro42:
Node 17 is End-of-Life since 2022-06-01. Still using it causes all sorts of logspam in the command line and in GitHub action logs, and risks undefined behavior.

Also, some actions were still using Node 16.


lucaswerkmeister (approved):
LGTM, but I just saw there’s also an `.nvmrc` with 16 in it – do you want to update that as well?

micgro42:
> LGTM, but I just saw there’s also an `.nvmrc` with 16 in it – do you want to update that as well?

Oh, I assumed that wouldn't exist because we are using sail everywhere. Will update it as well.

lucaswerkmeister:
I mean, I wouldn’t mind deleting it either, but I didn’t add it. As long as it has the same version as the docker-compose file and everything else, I guess the nvmrc doesn’t hurt.

micgro42:
> I mean, I wouldn’t mind deleting it either, but I didn’t add it. As long as it has the same version as the docker-compose file and everything else, I guess the nvmrc doesn’t hurt.

Yeah. Anyway, now it is correct. We can delete it later, if we want.

lucaswerkmeister (approved):

PR #452, opened 2022-10-28
Correctly send list parameters to the action API

lucaswerkmeister:
If any list items contain '|', we shouldn’t combine the list items with '|', but instead with U+001F (ASCII Unit Separator), with an additional U+001F at the beginning to mark the payload as using this separator instead of the usual pipe.


guergana (approved):
Nice to have this check for encoding in the url :+1:

lucaswerkmeister:
Oops, I think I merged this too early.

PR #454, opened 2022-10-28
Parse and format time values in mismatch results

lucaswerkmeister:
Add the formatted values as a separate `formatted` prop of the Results page, but fold it into the LabelledMismatch objects (wikidata_value member) in addLabels(), so that the “interior” components (MismatchesTable, MismatchRow) don’t need any changes.


resources/js/types/Mismatch.ts
```diff
@@ -27,6 +27,12 @@ interface Mismatch {
 
 export interface LabelledMismatch extends Mismatch {
     property_label: string,
+    /**
+     * A display label for the mismatch’s wikidata_value;
+     * this may be the label of the referenced item for an item-type property,
+     * or a formatted data value for other properties.
```
micgro42, line 33: > for other properties

Right now, only for time properties, right?

micgro42 (commented):
Looks good, just a single question about one phpdoc comment

resources/js/types/Mismatch.ts
lucaswerkmeister: Yes, though I don’t want to specify this here (to me, “this may be” already implies that the property is optional, without going into detail about when it exists or not).

micgro42: Mh, I find this to be ambiguous enough that I would like to see it changed.

How about this:

```suggestion
     * or a formatted data value for some other properties.
```

micgro42 (dismissed):

lucaswerkmeister:
Rebased again (GitHub reported a conflict, though locally the rebase was clean).

micgro42 (approved):
just a rebase

PR #455, opened 2022-10-31
Correctly send list parameters to the action API

lucaswerkmeister:
If any list items contain '|', we shouldn’t combine the list items with '|', but instead with U+001F (ASCII Unit Separator), with an additional U+001F at the beginning to mark the payload as using this separator instead of the usual pipe.

---

This pull request is a repeat of #452, which I merged too early. It’s based on the `parse+format-2` branch (#449), so that GitHub will only show the one new commit that actually belongs to this pull request; however, it should only be merged after that other branch / PR has been merged into `main`.

### DIFF
--- a/app/Services/WikibaseAPIClient.php
+++ b/app/Services/WikibaseAPIClient.php
@@ -27,6 +27,15 @@ class WikibaseAPIClient
         $this->cache = $cache;
     }
 
+    private function list(array $values): string
+    {
+        if (str_contains(implode($values), '|')) {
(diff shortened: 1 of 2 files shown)


micgro42 (dismissed):

lucaswerkmeister:
Meh, GitHub needed another force-push anyways to understand that the PR is only supposed to be about one commit :/

micgro42 (approved):
nothing changed since my last approval, I think

PR #456, opened 2022-10-31
Add meta_wikidata_value to DB and CSV

hasanakg:
Phabricator ticket: [T321165](https://phabricator.wikimedia.org/T321165)


database/factories/MismatchFactory.php
```diff
@@ -46,6 +46,7 @@ public function definition()
                 }
                 return $this->getRandomValueForDatatype(self::PROPERTIES[$propertyId]);
             },
+            'meta_wikidata_value' => 'Q' . $this->faker->optional(0.7)->randomNumber(),
```
lucaswerkmeister, line 49: I think this needs to depend on the `property_id` (like the `wikidata_value` above), otherwise it’ll generate meta values for non-time properties and cause validation failures. (I assume this is causing the CI errors.)

config/mismatches.php
```diff
@@ -17,6 +17,10 @@
         'wikidata_value' => [
             'max_length' => 1500 // Longest allowed value on wikidata
         ],
+        'meta_wikidata_value' => [
+            'max_length' => 100,
+            'format' => '/^Q\d+$/i'
```
lucaswerkmeister, line 22: Q0 shouldn’t be allowed; also, I don’t see why we would want to allow e.g. `q123`, so I’d remove the `i` flag as well.

Also, in `ItemId::PATTERN`, we terminate the pattern with `\z` instead of `$` , which [apparently](https://www.php.net/manual/en/regexp.reference.escape.php) means that it will only match at the end of the full input, even in multi-line mode. I’m not sure where this regex is evaluated, and with which flags, so I think it might be a good idea to use `\z` to make sure it only matches at the end of input. (I guess for the same reason we should also use `\A` instead of `^`? I don’t know why `ItemId::PATTERN` doesn’t do that, to be honest.)

docs/AdminGuide.md
```diff
@@ -100,6 +100,7 @@ Example entry:
     "mismatch_id": 1,
     "item_id": "Q3570615",
     "property_id": "P5474221",
+    "meta_wikidata_value": "Q71706 ",
```
lucaswerkmeister, line 103: There’s an extra space at the end here that probably shouldn’t be there.

tests/Feature/ValidateCSVTest.php
```diff
@@ -33,7 +33,7 @@ public function invalidLineProvider(): iterable
         yield 'missing statement guid' => [
             function (array $config): array {
                 return [
-                    ',P569,3 April 1934,1934-04-03,http://www.example.com', // Emulate missing guid
+                    ',P569,3 April 1934,,1934-04-03,http://www.example.com', // Emulate missing guid
```
lucaswerkmeister, line 36: Do we have some tests that leave out the new field? IIUC, it’s supposed to be optional – not only can you leave the column empty in your CSV, but not having the column at all is also supposed to be supported, I believe.

lucaswerkmeister (commented):
Looks pretty good; some minor comments before I try it out locally.

tests/Feature/ValidateCSVTest.php
lucaswerkmeister: I tried it out and it looks like the new column is actually required:

> CSV import parsing error at line 0: Unrecognized column headers, please include exactly the following column headers: statement_guid, property_id, wikidata_value, meta_wikidata_value, external_value, external_url

config/mismatches.php
hasanakg: If we change the `ItemId::PATTERN`, we will change the current validation and we have the same `/i` flag within `guid` field too. When `guid` is parsed, we use `strtoupper` to make store `item_id` correctly, but we don't do it on `guid` itself. We have different uses of these regular expressions. I'm changing this one as you suggest but I think we should talk about others as well.

lucaswerkmeister: Yeah, we should only change this one regex. Statement IDs can contain lowercase item IDs, so those should still be allowed.

app/Jobs/ValidateCSV.php
```diff
@@ -147,4 +157,22 @@ private function checkValueErrors($mismatch, WikidataValue $valueRule): ?string
 
         return null;
     }
+
+    private function checkMetaValueErrors($mismatch, MetaWikidataValue  $metaRule): ?string
+    {
+        $validator = Validator::make([
+            'meta_wikidata_value' => [
+                'property' => $mismatch['property_id'],
+                'meta_value' => $mismatch['meta_wikidata_value']
```
lucaswerkmeister, line 166: Can we call the array key here `'meta_wikidata_value'` as well? (Inside the validator, the variable is also called `$metaWikidataValue` again.)

app/Rules/MetaWikidataValue.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        [
+            'property' => $property,
+            'meta_value' => $metaWikidataValue
+        ] = $value;
+
+        try {
+            if (!empty($property) && !empty($metaWikidataValue)) {
+                $dataTypes = $this->wikidata->getPropertyDatatypes([$property]);
+                return array_key_exists($property, $dataTypes) && $dataTypes[$property] == 'time';
```
lucaswerkmeister, line 44: IMHO this `array_key_exists()` shouldn’t be here – `getPropertyDatatypes()` should guarantee that the property is included in the return value. (It looks like it currently doesn’t do that, so I’ll upload a pull request to fix that.)

lucaswerkmeister (commented):
Looks pretty good, two small comments. Haven’t tried it locally yet.

app/Rules/MetaWikidataValue.php
hasanakg: Actually that's why I had to add this check. Ping me when you open a PR for that then I will update this part

lucaswerkmeister (approved):

lucaswerkmeister:
We want to delay merging this until [allow reporting of mismatches with empty Wikidata value](https://phabricator.wikimedia.org/T313468) is also done, so that we can make the two breaking changes to the CSV format (add `meta_wikidata_value` and `item_id`) at the same time (more or less), and users don’t have to adjust their workflows twice.

guergana:
This was already approved, right @lucaswerkmeister ? I can re-approve it if it was.

lucaswerkmeister (approved):

lucaswerkmeister:
Well, we still don’t want to merge it until #473 is also ready.

guergana:
> Well, we still don’t want to merge it until #473 is also ready.

oh, i didn't know you could approve if you worked on it :D

PR #457, opened 2022-11-01
Update AdminGuide with intructions to change import expiry dates

guergana:
Bug: [T321586](https://phabricator.wikimedia.org/T321586)

### DIFF
--- a/docs/AdminGuide.md
+++ b/docs/AdminGuide.md
@@ -123,11 +123,6 @@ Example entry:
     become mismatch-finder
     ```
 
-1. Get into the database using the toolforge alias for accessing Mariadb 
-    ```bash
-    sql tools
-    ```
-
 1. Get into the `mismatch-finder-repo` folder
 
     ```bash
--- a/docs/AdminGuide.md
+++ b/docs/AdminGuide.md
@@ -128,10 +128,15 @@ Example entry:
     sql tools
     ```
 
-1. Use the Wikidata mismatch finder database
+1. Get into the `mismatch-finder-repo` folder
 
-    ```sql
-    USE s54796__mismatch_finder; 
+    ```bash
+    cd mismatch-finder-repo
+
+1. Go into the project's database using artisan
+
+    ```bash
+    php artisan db
     ```
 
 1. Now inside the database we can perform the update operation. 
--- a/docs/AdminGuide.md
+++ b/docs/AdminGuide.md
@@ -108,3 +108,54 @@ Example entry:
     "time": "2021-10-05 14:44:59",
 }
 ```
+
+## Update the expiry date of mismatches
+
+1. SSH into toolforge.
+
+    ```bash
+    ssh <your_username>@login.toolforge.org
+    ```
+
+1. The `become` commands allows you to sudo to another user while retaining your personalized environment.
+
+    ```bash
+    become mismatch-finder
+    ```
+
+1. Get into the database using the toolforge alias for accessing Mariadb 
+    ```bash
+    sql tools
+    ```
+
+1. Use the Wikidata mismatch finder database
+
+    ```bash
+    USE s54796__mismatch_finder; 
+    ```
+
+1. Now inside the database we can perform the update operation. 
+
+    ```bash
+    UPDATE import_meta SET expires='<date_of_expiration>' where id = <id_of_entry_to_update>; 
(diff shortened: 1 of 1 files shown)


docs/AdminGuide.md
```diff
@@ -108,3 +108,54 @@ Example entry:
...
+    ```bash
+    sql tools
+    ```
+
+1. Use the Wikidata mismatch finder database
+
+    ```bash
+    USE s54796__mismatch_finder; 
+    ```
+
```
hasanakg, line 136: I'm not sure if the prefix (`s54796`) is same for every mismatch finder instance, so it'd be great to run `SHOW DATABASES;` command to be able to pick the correct DB.

docs/AdminGuide.md
```diff
@@ -108,3 +108,54 @@ Example entry:
...
+    ```bash
+    UPDATE import_meta SET expires='<date_of_expiration>' where id = <id_of_entry_to_update>; 
+    ```
+
+    Example:
+
+    ```bash
+    UPDATE import_meta SET expires='2022-11-01' where id = 6
+    ```
```
hasanakg, line 147: I know these are run in bash but using `sql` as code block language make them prettier imho.
For example:
```sql
UPDATE import_meta SET expires='<date_of_expiration>' where id = <id_of_entry_to_update>; 
```

docs/AdminGuide.md
guergana: The database is indeed always the same. It just got that suffix number assigned automatically when it was created.

lucaswerkmeister: The database will always be the same on Toolforge, but I think Hasan was thinking about other installations of Mismatch Finder? (E.g. our local installs.)

guergana: It can be rewritten to `cd mismatch-finder-repo` and then `php artisan db` as @lucaswerkmeister mentions but I don't that's any better tbh. The db name is not changing. Toolforge just created this db since the start. This would be useful if there are plans to move the database at some point.

docs/AdminGuide.md
```diff
@@ -108,3 +108,54 @@ Example entry:
...
+1. Use the Wikidata mismatch finder database
+
+    ```sql
+    USE s54796__mismatch_finder; 
+    ```
+
+1. Now inside the database we can perform the update operation. 
+
+    ```sql
+    UPDATE import_meta SET expires='<date_of_expiration>' where id = <id_of_entry_to_update>;
```
mariushoch, line 140: Shall we maybe advice users to do this in a transaction? Otherwise for example a forgotten `WHERE` can hurt pretty bad.

guergana: What would be the difference @mariushoch ? I am familiar with transactions.

mariushoch: Changes you do in a transaction can be rolled back, so if you for example run `UPDATE import_meta SET expires='<date_of_expiration>';` (missing `WHERE` clause, will set the new expiration date for ALL imports), you can just issue a `ROLLBACK` to revert your mistake. Otherwise the only option would be to restore the old contents from a backup.

guergana: that sounds definitely like a good idea. so the new command would be like this?

```sql
START TRANSACTION;
UPDATE import_meta SET expires='<date_of_expiration>' where id = <id_of_entry_to_update>; 
COMMIT;
```

guergana: otherwise you can add a suggestion to the PR and i commit it. @mariushoch

mariushoch: Yes, that looks good to me :)

guergana: Done!

docs/AdminGuide.md
```diff
@@ -108,3 +108,63 @@ Example entry:
...
+    ssh <your_username>@login.toolforge.org
+    ```
+
+1. The `become` commands allows you to sudo to another user while retaining your personalized environment.
+
+    ```bash
+    become mismatch-finder
+    ```
+
+1. Get into the database using the toolforge alias for accessing Mariadb
```
lucaswerkmeister, line 126: This step should be removed now (it was replaced by the next two steps).

guergana: true!

PR #459, opened 2022-11-07
Make tests more resilient by reusing user in db

micgro42:
The hope is, that by not recreating a new user for each test, the loginAs method will work more robustly.

Bug: T311506


lucaswerkmeister (approved):
let’s try it

guergana:
Cool! Let's see how this approach works :) Thanks.

PR #470, opened 2022-11-10
Fix missing labels on Items without mismatches

micgro42:
This was probably introduced in #444, when overlooking to consider that $mismatches does not contain info about Items that do not have mismatches and so fails to retrieve their labels.

TODO:
- [x] actually try it out
- [x] tests
- [x] better commit message

Bug: T321977


micgro42:
Would be nice if someone reviewing could also actually try it out locally. My local installation still doesn't work correctly.

guergana (approved):
Works like a charm. Thx.

PR #472, opened 2022-11-15
meta_wikidata_value to replace calendar model if exists

chukarave:
check if meta_wikidata_value exists and if so
append it to the base URL and assign that as
the calendar model of the parsed time values.

Bug: T321174

### DIFF
--- a/app/Http/Controllers/ResultsController.php
+++ b/app/Http/Controllers/ResultsController.php
@@ -59,12 +59,26 @@ class ResultsController extends Controller
 
         $timeValues = $this->extractTimeValues($mismatches, $datatypes);
         $parsedTimeValues = $wikidata->parseValues($timeValues);
-        // assert if calendar model is specified
-        if (!empty($mismatches->meta_wikidata_value)) {
-            $parsedTimeValues[ 'calendarmodel' ] =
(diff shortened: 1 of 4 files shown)


lucaswerkmeister (commented):
Can we extract the code in `ResultsController` into a separate method? So the chain might be something like:

```php
$timeValues = $this->extractTimeValues($mismatches, $datatypes);
$parsedTimeValues = $wikidata->parseValues($timeValues);
$updatedTimeValues = $this->updateTimeValues($mismatches, $parsedTimeValues);
$formattedTimeValues = $wikidata->formatValues($updatedTimeValues, $lang);
```

There are also two `var_dump()`s left over. Also, I’m not sure “assert” is the right word – maybe “check”?

lucaswerkmeister (approved):
LGTM; I tried to test it locally by adding `options={"showcalendar": "auto"}` to the `wbformatvalue` API call, but of course it doesn’t work because that API call goes to production Wikidata, where support for that option hasn’t been deployed yet (so you just see the timestamp without calendar model). I think we can just merge this, test it next week when the train has reached Wikidata, and potentially add any necessary fixes then.

PR #473, opened 2022-11-16
Add item_id column to CSV file format

lucaswerkmeister:
This is now a required column of the CSV file format. The intention is to make the `statement_guid` column optional later, but for now it is still required; a custom validator ensures that the item IDs in both are consistent.

Bug: T323203


guergana:
In this case the [exampleMismatchFile.csv](https://github.com/wmde/wikidata-mismatch-finder/blob/main/docs/exampleMismatchFile.csv) file needs to be updated as well

guergana (changes requested):
In this case the [exampleMismatchFile.csv](https://github.com/wmde/wikidata-mismatch-finder/blob/main/docs/exampleMismatchFile.csv) file needs to be updated as well

docs/UserGuide.md
```diff
@@ -60,7 +61,7 @@ statement_guid,property_id,wikidata_value,meta_wikidata_value,external_value,ext
 
 _**Note**: The data `wikidata_value`, `external_value`, `external_url` should be limited to a length of 1500 characters maximum._
 
-Additionally, each row of the csv file must contain exactly 4 commas (`,`). Optional values can simply be left out.
+All columns must be present; optional values can be left empty.
 
 You can consult this [mismatch file](exampleMismatchFile.csv) for a valid example.
```
guergana, line 66: This file mentioned here needs to be updated to be consistent with the new changes

lucaswerkmeister: Ok, then I’ll push a commit to #456 to update the file there as well. (Here I was mostly following the list of files changed there.)

guergana: Thanks. That's a good idea.

docs/UserGuide.md
```diff
@@ -60,7 +61,7 @@ statement_guid,property_id,wikidata_value,meta_wikidata_value,external_value,ext
 
 _**Note**: The data `wikidata_value`, `external_value`, `external_url` should be limited to a length of 1500 characters maximum._
 
-Additionally, each row of the csv file must contain exactly 4 commas (`,`). Optional values can simply be left out.
+All columns must be present; optional values can be left empty.
```
guergana, line 64: Mentioning that leaving empty is putting commas is worth it here. I am not sure if all users will be familiar with how empty fields work in csv. :thinking:

lucaswerkmeister: Maybe, but I’m not sure how to phrase it… I guess we can use a specific column as an example?

> , e.g. `,,` for empty `meta_wikidata_value`.

lucaswerkmeister: Done.

guergana (approved):

guergana:
LGTM

PR #474, opened 2022-11-17
Add "missing" review_status

lucaswerkmeister:
The code part of this is mostly straightforward – I mainly searched for occurrences of “both” to find all references to mismatch statuses and added “missing” directly after the “wikidata” option.

The migration is more painful. Laravel doesn’t support altering ENUM columns at all. In MySQL, we can issue a raw ALTER TABLE command, adding the new enum value (at the end of the enum, which is efficient since MySQL 5.1); in SQLite there’s nothing equivalent (it’s apparently just a string under the hood? but we still need to tell Laravel that the column now has an extra supported value), but we only support SQLite for testing, so just drop the entire table and recreate it from scratch.

Bug: T323204


docs/UserGuide.md
```diff
@@ -147,6 +147,7 @@ The only possible values for a review status are:
 - `"wikidata"` - The mismatching information is on Wikidata.
```
micgro42, line 147: Not caused by this PR, but I find this line highly confusing. "mismatching" in this context mean "correct" or "incorrect"?

database/migrations/2022_11_17_153635_add_missing_to_mismatches_review_status.php
```diff
@@ -0,0 +1,114 @@
...
+            $table->string('external_value', 2048); // Arbitrary length above 1500 chars, upper limit for external value
+            $table->string('external_url', 2048)->nullable(); // Arbitrary length above 1500 chars, upper limit for url
+            $table->enum('review_status', [
+                'pending',
+                'wikidata',
+                'external',
+                'both',
+                'none',
+                'missing'
+            ])->default('pending');
```
micgro42, line 64: I think that explicit default is new compared to https://github.com/wmde/wikidata-mismatch-finder/blob/main/database/migrations/2021_08_04_102448_recreate_mismatches_table_with_item_id.php ? But makes sense to me, not something to change

micgro42 (commented):
Still have to try it out locally, but looks good overall

database/migrations/2022_11_17_153635_add_missing_to_mismatches_review_status.php
lucaswerkmeister: I mainly copied this from https://github.com/wmde/wikidata-mismatch-finder/blob/main/database/migrations/2021_08_18_121955_rename_mismatch_status.php (looks like I forgot to mention that in the commit message), where the default was already there.

lucaswerkmeister: d’oh, and “I copied this from $oldMigration” of course means that the `meta_wikidata_value` is missing, which is why so many tests are failing :facepalm:

micgro42:
I tried it out and I think the table now looks the way it is supposed to look :+1:

docs/UserGuide.md
```diff
@@ -147,6 +147,7 @@ The only possible values for a review status are:
...
 - `"external"` - The mismatching information is on the external source.
 - `"both"` - Both sources are incorrect.
+- `"missing"` - The information is only available in the external source and missing on Wikidata.
```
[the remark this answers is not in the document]
lucaswerkmeister: Done.

app/Services/ImportReport.php
```diff
@@ -36,6 +36,7 @@ public function generateCSV(string $path): void
                 $each_import->status,
                 $mismatches_count,
                 $mismatches_wikidata,
+                $mismatches_missing,
```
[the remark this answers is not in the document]
lucaswerkmeister: Yeah, I think not including this change in the PR would’ve been more cumbersome, since the header row includes all the accepted statuses (I found this file via a failing test).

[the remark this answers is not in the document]
micgro42: makes sense :+1:

database/migrations/2022_11_17_153635_add_missing_to_mismatches_review_status.php
micgro42: seems to work now :+1:

micgro42 (approved):
Looks good to me!

PR #475, opened 2022-11-17
Show missing wikidata value as "None" and adjust validation

guergana:
Bug: T323206

Todo - tests: 
 - [x] check that the 'None' value is correctly added when statement-guid and wikidata value are missing. 
 - [x] check that statement-guid cannot be present without wikidata-value and viceversa (2 tests)

### DIFF
--- a/app/Http/Controllers/ResultsController.php
+++ b/app/Http/Controllers/ResultsController.php
@@ -66,7 +66,8 @@ class ResultsController extends Controller
                 'user' => $user,
                 'item_ids' => $requestedItemIds,
                 // Use wikidata to fetch labels for found entity ids
-                'labels' => $wikidata->getLabels($entityIds, $lang),
+                // array_filter removes empty values from an array if no callback argument is passed
+                'labels' => $wikidata->getLabels(array_filter($entityIds), $lang),
                 'formatted_values' => $formattedTimeValues,
             ],
             // only add 'results' prop if mismatches have been found
--- a/app/Http/Controllers/ResultsController.php
+++ b/app/Http/Controllers/ResultsController.php
@@ -94,10 +94,12 @@ class ResultsController extends Controller
         $idsAsKeys = [];
 
         foreach ($mismatches as $mismatch) {
-            $idsAsKeys[$mismatch->item_id] = null;
+            if ($mismatch->wikidata_value !== '' && $mismatch->statement_guid !== '') {
+                $idsAsKeys[$mismatch->item_id] = null;
 
-            if ($datatypes[$mismatch->property_id] === 'wikibase-item') {
-                $idsAsKeys[$mismatch->wikidata_value] = null;
+                if ($datatypes[$mismatch->property_id] === 'wikibase-item') {
+                    $idsAsKeys[$mismatch->wikidata_value] = null;
+                }
             }
         }
 
--- a/app/Http/Controllers/ResultsController.php
+++ b/app/Http/Controllers/ResultsController.php
@@ -94,9 +94,9 @@ class ResultsController extends Controller
         $idsAsKeys = [];
 
         foreach ($mismatches as $mismatch) {
-            if ($mismatch->wikidata_value !== '' && $mismatch->statement_guid !== '') {
-                $idsAsKeys[$mismatch->item_id] = null;
-
+            $idsAsKeys[$mismatch->item_id] = null;
+            
+            if ($mismatch->wikidata_value !== '') {
                 if ($datatypes[$mismatch->property_id] === 'wikibase-item') {
                     $idsAsKeys[$mismatch->wikidata_value] = null;
                 }
--- a/app/Http/Controllers/ResultsController.php
+++ b/app/Http/Controllers/ResultsController.php
@@ -66,7 +66,6 @@ class ResultsController extends Controller
                 'user' => $user,
                 'item_ids' => $requestedItemIds,
                 // Use wikidata to fetch labels for found entity ids
-                // array_filter removes empty values from an array if no callback argument is passed
                 'labels' => $wikidata->getLabels(array_filter($entityIds), $lang),
                 'formatted_values' => $formattedTimeValues,
             ],
--- a/app/Http/Controllers/ResultsController.php
+++ b/app/Http/Controllers/ResultsController.php
@@ -66,7 +66,7 @@ class ResultsController extends Controller
                 'user' => $user,
                 'item_ids' => $requestedItemIds,
                 // Use wikidata to fetch labels for found entity ids
-                'labels' => $wikidata->getLabels(array_filter($entityIds), $lang),
+                'labels' => $wikidata->getLabels($entityIds, $lang),
                 'formatted_values' => $formattedTimeValues,
             ],
             // only add 'results' prop if mismatches have been found
--- a/app/Jobs/ValidateCSV.php
+++ b/app/Jobs/ValidateCSV.php
@@ -111,7 +111,7 @@ class ValidateCSV implements ShouldQueue
                 'regex:' . $rules['item_id']['format']
             ],
             'statement_guid' => [
-                'required',
+                'required_with:wikidata_value',
                 'max:' . $rules['guid']['max_length'],
                 'regex:' . $rules['guid']['format']
             ],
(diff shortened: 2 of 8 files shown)


app/Http/Controllers/ResultsController.php
```diff
@@ -66,7 +66,8 @@ public function index(MismatchGetRequest $request, WikibaseAPIClient $wikidata):
                 'user' => $user,
                 'item_ids' => $requestedItemIds,
                 // Use wikidata to fetch labels for found entity ids
-                'labels' => $wikidata->getLabels($entityIds, $lang),
+                // array_filter removes empty values from an array if no callback argument is passed
```
micgro42, line 69: This comment should probably not be here, it adds no additional information and is prone to get outdated quickly.

(I'd probably also argue against the comment line directly above, but that was not added by this PR, so that's out of scope here.)

app/Http/Controllers/ResultsController.php
```diff
@@ -66,7 +66,8 @@ public function index(MismatchGetRequest $request, WikibaseAPIClient $wikidata):
...
+                'labels' => $wikidata->getLabels(array_filter($entityIds), $lang),
```
micgro42, line 70: Under what circumstances would there be an empty element in `$entityIds`?

resources/js/Components/MismatchRow.vue
```diff
@@ -9,7 +9,11 @@
             </wikit-link>
         </td>
         <td :data-header="$i18n('column-wikidata-value')">
-            <wikit-link class="break-line-link" :href="statementUrl" target="_blank">
+            <span class="empty-value" v-if="!mismatch.value_label && !mismatch.wikidata_value">
```
micgro42, line 12: We probably want to ensure that `mismatch.wikidata_value` is `null` and test for that here specifically, so that we can differentiate it from a value that happens to be, for example, `0`.

tests/Feature/ValidateCSVTest.php
```diff
@@ -71,18 +71,6 @@ function (array $config): array {
             }
         ];
 
-        yield 'missing statement guid' => [
```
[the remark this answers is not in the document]
guergana: not really. because the missing statement guid shouldn't fail. i have to write new tests.

app/Http/Controllers/ResultsController.php
guergana: in the case of the wikidata_value being empty. otherwise `getLabels` calls the WikibaseAPIClient with an invalid id, and it gives an error. you can try it out.

resources/js/Components/MismatchRow.vue
guergana: it's not null, it's an empty string.

tests/Feature/ValidateCSVTest.php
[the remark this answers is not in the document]
micgro42: But only when the statement guid _AND_ the wikidata value are missing, right? If one is missing and the other one is there, then that should indeed be an error.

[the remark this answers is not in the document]
guergana: yes, exactly, but it's a different test, which i should write still as i indicated in the description.

app/Http/Controllers/ResultsController.php
micgro42: `array_filter` is well documented at the PHP manual and can easily be looked up there: https://www.php.net/manual/en/function.array-filter.php

The trouble with such comments is that they are duplicating what is already written in the code itself and tend to get overlooked and out of date when someone changes the code.

Here is a bit more about how to fix such an issue: https://refactoring.guru/smells/comments

Though in this case, I would still argue that the comment can just be dropped, because it merely explains the default behavior of a core PHP function.

app/Http/Controllers/ResultsController.php
micgro42: From reading the code more, it seems the only way that an empty value can end up in `$entityIds` is if it is in the array returned by `extractItemIds`. Which in turn means we may want to adjust the condition in `extractItemIds` to actually only return item IDs:

```php
    private function extractItemIds(LazyCollection $mismatches, array $datatypes): array
    {
        return array_keys($mismatches->reduce(
            function (array $ids, Mismatch $mismatch) use ($datatypes) {
                $ids[$mismatch->item_id] = null;
                if ($datatypes[$mismatch->property_id] === 'wikibase-item') {
                    $ids[$mismatch->wikidata_value] = null;
                }
                return $ids;
            },
            []
        ));
    }
```

app/Http/Controllers/ResultsController.php
guergana: I thought it might help, but if it doesn't then yes. :) Removed.

app/Http/Controllers/ResultsController.php
guergana: I agree, it's better.

app/Http/Controllers/ResultsController.php
```diff
@@ -94,10 +94,12 @@ private function extractItemIds(LazyCollection $mismatches, array $datatypes): a
         $idsAsKeys = [];
 
         foreach ($mismatches as $mismatch) {
-            $idsAsKeys[$mismatch->item_id] = null;
+            if ($mismatch->wikidata_value !== '' && $mismatch->statement_guid !== '') {
```
lucaswerkmeister, line 97: I think we still need the item label even if the value is missing, right? The condition should probably be only on the inner if. (Also, I find it confusing to test the `statement_guid` here – `wikidata_value` should be the only thing we care about IMHO.)

tests/Feature/ValidateCSVTest.php
```diff
@@ -163,24 +164,25 @@ function (array $config): array {
             }
         ];
 
-        yield 'missing wikidata value' => [
+        yield 'invalid meta wikidata value' => [
```
lucaswerkmeister, line 167: IIUC, this is now the same three test cases as before (except that the validation error changes a bit), but “missing wikidata value [when statement guid is present]” and “invalid meta wikidata value” have swapped places, which makes the diff more confusing IMHO… can we swap them back?

tests/Feature/WebResultsRouteTest.php
```diff
@@ -118,8 +118,6 @@ public function test_results_route_retrieves_mismatches()
                 $itemMismatch->property_id,
                 $stringMismatchQid,
                 $stringMismatch->property_id,
-                // as well as item id of an item mismatch value
-                $itemMismatch->wikidata_value,
```
lucaswerkmeister, line 122: Does this have to be removed? I don’t see any changes to the `MismatchFactory` in this PR, so the `wikidata_value` in this test should still always exist, right?

app/Jobs/ValidateCSV.php
```diff
@@ -111,7 +111,7 @@ private function checkFieldErrors($mismatch): ?string
                 'regex:' . $rules['item_id']['format']
             ],
             'statement_guid' => [
-                'required',
+                'required_with:wikidata_value',
```
lucaswerkmeister, line 114: Nice and compact :+1:

micgro42:
I'm surprised that tests/Feature/ValidateCSVTest.php does not seem to have any tests for _valid_ CSV, but that's out of scope of this PR.
I think all my comments have been resolved.

guergana:
> I'm surprised that tests/Feature/ValidateCSVTest.php does not seem to have any tests for _valid_ CSV, but that's out of scope of this PR.
> I think all my comments have been resolved.

Yes i also was looking for the correct import but didnt manage to formulate it, just thought it must be ok if we check for all the possible fails 😅 We could write a ticket for that.

lucaswerkmeister:
Diff looks nice and clean to me now; I’ll try it out locally.

lucaswerkmeister (changes requested):
In the validation, `meta_wikidata_value` should probably be `prohibited_unless:wikidata_value`, right? We don’t want the `meta_wikidata_value` to be set if the `wikidata_value` is empty.

lucaswerkmeister:
Looks good to me apart from the one comment I left above.
image

lucaswerkmeister (approved):

PR #476, opened 2022-11-21
Refactor extractItemIds to make it easier to understand

micgro42:
This basically extracts the reduce callback into its own variable in order to be able to name it, and make what is happening more explicit.

Also, strategically adds two newlines inside that callback to emphasize what logic directly belongs together.

### DIFF
--- a/app/Http/Controllers/ResultsController.php
+++ b/app/Http/Controllers/ResultsController.php
@@ -91,16 +91,22 @@ class ResultsController extends Controller
 
     private function extractItemIds(LazyCollection $mismatches, array $datatypes): array
     {
-        return array_keys($mismatches->reduce(
-            function (array $ids, Mismatch $mismatch) use ($datatypes) {
-                $ids[$mismatch->item_id] = null;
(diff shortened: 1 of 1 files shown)


guergana (commented):
Maybe the `reduce` is so that repeated values in the function are filtered out? Anyway i don't think extracting the same function as a variable makes it more understandable. i would still consider it more useful to add a comment and explain what `reduce` is doing specifically in that case because Larave's  official examples in the docs are about using it to add values within an array. And this is not what we are doing.

lucaswerkmeister:
We could probably also turn this into a normal loop – when I wrote this code, I wasn’t sure how much to stay with these “collection” objects and their methods, IIRC.

guergana:
> 

Could you make a code suggestion on how to turn it into a normal loop? Do you mean with a `foreach` or `map`?

guergana:
I guess then we can close this PR, as the issue has been worked on in #477

guergana:
@micgro42 given the fact that we write open source code, I think we should strive for simple code when possible.

micgro42:
Nothing wrong with that.

PR #477, opened 2022-11-22
Refactor confusing ResultsController methods

lucaswerkmeister:
It turns out a `LazyCollection` is iterable, so we can skip the confusing `reduce()` call and use a conventional `foreach` loop instead.

Also rename the `$ids` variables to make it clear that they include the IDs as keys (we use this trick to automatically eliminate duplicates).


guergana (approved):
It looks better now. Thanks!

micgro42 (approved):
This is one way to make it easier to read. Thank you

PR #479, opened 2022-11-25
Add user agent to server-side Wikidata requests

micgro42:
A user-agent is required following the User Agent policy: https://meta.wikimedia.org/wiki/User-Agent_policy

Bug: T321835

### DIFF
--- a/app/Services/WikibaseAPIClient.php
+++ b/app/Services/WikibaseAPIClient.php
@@ -39,6 +39,7 @@ class WikibaseAPIClient
     private function get(string $action, array $params): Response
     {
         $response = Http::withMiddleware($this->cache)
+            ->withHeaders([ 'User-Agent' => $this->getUserAgent() ])
             ->get($this->baseUrl, array_merge([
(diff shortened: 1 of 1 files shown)


app/Services/WikibaseAPIClient.php
```diff
@@ -54,6 +55,27 @@ private function get(string $action, array $params): Response
         return $response;
     }
 
+    /**
+     * @see https://meta.wikimedia.org/wiki/User-Agent_policy
+     */
+    private function getUserAgent(): string
+    {
+        $clientName = 'Wikidata MismatchFinder';
```
lucaswerkmeister, line 63: I think this should be spelled without a space – otherwise it means a product named “Wikidata” followed by another (less specific) product “MismatchFinder”. I suggest `Wikidata-Mismatch-Finder`.

micgro42: Done.

lucaswerkmeister (approved):
Thanks!

PR #485, opened 2022-11-28
Replace heading with logo and link to index page

lucaswerkmeister:
Bug: T323680


resources/js/Pages/Layout.vue
```diff
@@ -2,12 +2,14 @@
     <div class="website">
         <main class="content-wrap">
             <header>
-                <img src="/images/wikidata-logo.svg" class="wikidata-logo" alt="Wikidata-logo" width="160" />
+                <InertiaLink href="/">
+                    <img src="/images/mismatch-finder-logo.svg" class="wikidata-logo" alt="" width="384" />
+                </InertiaLink>
```
micgro42, line 7: Now that link does not have any content that a screen-reader could read out. (I'm not yet fully sure how to fix this, but it is not great.)

I would say we could use `$i18n('mismatch-finder-title')` as the link text, and something more specific for this results page.

public/mix-manifest.json
```diff
@@ -3,5 +3,6 @@
     "/css/app.css": "/css/app.css",
     "/images/download-stats.svg": "/images/download-stats.svg",
     "/images/user.svg": "/images/user.svg",
-    "/images/wikidata-logo.svg": "/images/wikidata-logo.svg"
+    "/images/wikidata-logo.svg": "/images/wikidata-logo.svg",
```
micgro42, line 6: I think we can delete this line, since we deleted the file, right?

lucaswerkmeister: Indeed, done.

resources/js/Pages/Layout.vue
lucaswerkmeister: Hm, what if we move the heading inside the link?

```diff
diff --git a/resources/js/Pages/Layout.vue b/resources/js/Pages/Layout.vue
index 217bd53bd3..feb61399c5 100644
--- a/resources/js/Pages/Layout.vue
+++ b/resources/js/Pages/Layout.vue
@@ -4,8 +4,8 @@
             <header>
                 <InertiaLink href="/">
                     <img src="/images/mismatch-finder-logo.svg" class="wikidata-logo" alt="" width="384" />
+                    <h1 class="visually-hidden">{{ $i18n('mismatch-finder-title') }}</h1>
                 </InertiaLink>
-                <h1 class="visually-hidden">{{ $i18n('mismatch-finder-title') }}</h1>
                 <div class="auth-widget">
                     <auth-widget :user="user" />
                 </div>
```

I’m also not sure what you mean about the results page…

lucaswerkmeister: > (That they are not actually two different pages in the browser's history is something that I would consider a bug. Even the path is different (`/` vs `/results`).)

Fixing that in #486.

Are you suggesting to change the heading depending on which page the user is on (similar to the `<title>`)? That sounds doable, but I didn’t think it was part of the task at all (which is why I was so confused ^^).

micgro42: _I_ think those two page should have distinct headings, like "Mismatch Finder Results" and "Mismatch Finder Main Page" (or just "Mismatch Finder"), but I acknowledge that this is not part of this task. I'm fine with moving forward with what you suggested for this PR.

lucaswerkmeister:
Rebased onto main with all the fixups squashed into their respective commits.

micgro42 (approved):
Now the accessibly properties of that link look somewhat reasonable. Thanks!

PR #487, opened 2022-11-29
Update vertical spacing

lucaswerkmeister:
Remove the breakpoint-dependent spacing of the header, and instead always make it $dimension-layout-small. I can’t find any reason for the margin depending on the background (it was added in aee8dc75cfe), nor a record of Design asking for this, and we don’t want it now.

Add a margin to the top of the “back” button on the results page; this makes its total spacing (including the header margin) match that of the first heading on the home page (as headings have a default top margin). Also add a margin to the bottom of the “back” button, so it’s not too close to the description section (which it’s not directly related to).

Remove the top margin of the “which items should be checked?” heading on the home page, so that the total space between it and the preceding section is just the section’s own margin, matching the space between sections on the results page.

Bug: T323678

### DIFF
--- a/resources/js/Pages/Home.vue
+++ b/resources/js/Pages/Home.vue
@@ -235,6 +235,11 @@
     display: flex;
     justify-content: space-between;
     align-items: center;
+
+    h2 {
+        // the previous section already has enough margin-bottom
+        margin-top: 0;
+    }
 }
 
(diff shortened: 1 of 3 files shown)


sai-san (approved):
All specs are met. Thank you! 💯

PR #488, opened 2022-11-30
Update jest dependencies

chukarave:
Bug: T323900

PR #489, opened 2022-11-30
Update eslint dependencies

chukarave:
Merge after #488 

Bug: [T323900](https://phabricator.wikimedia.org/T323900)


package.json
```diff
@@ -16,15 +16,15 @@
     "devDependencies": {
         "@types/jest": "^29.2.3",
         "@types/lodash": "^4.14.184",
-        "@typescript-eslint/eslint-plugin": "^5.36.2",
-        "@typescript-eslint/parser": "^5.36.2",
+        "@typescript-eslint/eslint-plugin": "^5.45.0",
+        "@typescript-eslint/parser": "^5.45.0",
         "@vue/eslint-config-typescript": "^11.0.0",
```
lucaswerkmeister, line 21: Apparently there’s a version 11.0.2 of this now, so we should probably update it as well?

chukarave: I left it out cause i assumed we are avoiding updating vue related packages. i guess that can't hurt

lucaswerkmeister: Ah, I see… I think we can count this one as close to eslint, though.

lucaswerkmeister (approved):

PR #508, opened 2022-12-02
Fix app env in user agent

lucaswerkmeister:
The `'env'` config is defined in `config/app.php`, so the full config key is actually `'app.env'`.

Bug: T321835


micgro42 (approved):
Thanks for fixing this!

PR #510, opened 2022-12-14
ImportReport: Optimize database queries

lucaswerkmeister:
Previously, we were loading the full data of every mismatch in the mismatches table, and counting the mismatches per review status in memory. Also, we were querying the mismatches one import at a time, burdening the databases with a lot of random accesses (based on the index that exists on the import_id due to the foreign ID constraint) when really we want the data of all mismatches in the end.

Instead, have the database group and count the mismatches by import and status, and only send those counts over the wire. This should already be more efficient (only one full table scan, and less data transferred); it can optionally be optimized further later, by adding a covering index for the import_id and review_status, if we need it.

----

Not directly related to any Phabricator task; I’d noticed it a while ago while working on T323204, and now remembered it again because of T304794.

I’ve tested locally that http://localhost:1337/store/imports-overview produces the same output with and without this change. (There’s also a noticeable performance difference, from ca. 0.23 s down to 0.12 s. On Toolforge the CSV currently takes over three seconds to generate; I predict this will go down to less than half a second.)

### DIFF
--- a/app/Services/ImportReport.php
+++ b/app/Services/ImportReport.php
@@ -2,8 +2,8 @@
 
 namespace App\Services;
 
-use App\Models\Mismatch;
 use App\Models\ImportMeta;
+use Illuminate\Support\Facades\DB;
 
 class ImportReport
 {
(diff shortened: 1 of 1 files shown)


app/Services/ImportReport.php
```diff
@@ -17,14 +17,26 @@ public function generateCSV(string $path): void
...
+            ->get();
+
+        $review_statuses = config('mismatches.validation.review_status.accepted_values');
+
         foreach ($imports as $each_import) {
-            $mismatches = Mismatch::where('import_id', $each_import->id)->get();
-            $mismatches_count = $mismatches->count();
-            $review_statuses = config('mismatches.validation.review_status.accepted_values');
+            foreach ($review_statuses as $status) {
+                ${"mismatches_" . $status} = 0;
```
micgro42, line 30: Not sure about how happy I am about this syntax, over just using an associative array. But not introduced with this PR, so I don't consider this blocking.

micgro42: In particular, I was searching for a while for where `$mismatches_pending` is defined/set 🙄

micgro42 (approved):
Looks plausible to me, but I haven't tried it out. Is this meaningfully covered by tests?

app/Services/ImportReport.php
lucaswerkmeister: Yeah, I had the same issue when I was working on `missing` support ^^ happy to upload a follow-up change that uses an array instead.

lucaswerkmeister:
> Is this meaningfully covered by tests?

`ImportReportTest.php` looks like a reasonable test to me, yeah.

lucaswerkmeister:
> > Is this meaningfully covered by tests?
> 
> `ImportReportTest.php` looks like a reasonable test to me, yeah.

On second thought, there is a relevant case that’s not covered yet; see #511. (I’ll rebase this PR once that one is merged.)

lucaswerkmeister:
Rebased onto #511.

app/Services/ImportReport.php
micgro42: yep, let's do that in a follow-up

micgro42 (approved):

lucaswerkmeister:
Time to generate a report before this change: **3.08 seconds** (best of 10).

```console
$ (export LC_ALL=C; min=100; for i in {1..10}; do time=$(\time -f%e curl -so/dev/null https://mismatch-finder.toolforge.org/store/imports-overview 2>&1); if (( $(bc <<< "$time < $min") )); then min=$time; fi; done; printf '%s\n' "$min";)
3.08
```

After this change: **0.49 seconds** (ditto).

```console
$ (export LC_ALL=C; min=100; for i in {1..10}; do time=$(\time -f%e curl -so/dev/null https://mismatch-finder.toolforge.org/store/imports-overview 2>&1); if (( $(bc <<< "$time < $min") )); then min=$time; fi; done; printf '%s\n' "$min";)
0.49
```

just *barely* within my prediction lol

PR #513, opened 2022-12-19
Add new route for import result csv

chukarave:
This pr adds the route and introduces the
`ImportsResultsController`, which uses the new `ImportResults` service to generate the CSV dynamically. 
This also adds a new config section with the file name template and the headers.

Bug: T325158

### DIFF
--- a/app/Http/Controllers/ImportsResultsController.php
+++ b/app/Http/Controllers/ImportsResultsController.php
@@ -2,20 +2,23 @@
 
 namespace App\Http\Controllers;
 
+use App\Models\ImportMeta;
 use App\Services\ImportResults;
 
 class ImportsResultsController extends Controller
 {
     public function showResultsCsv(int $import_id, ImportResults $results)
     {
+        $import = ImportMeta::findOrFail($import_id);
+
         $filename = strtr(config('imports.results.filename_template'), [
             ':datetime' => now()->format('Y-m-dTH:i:s')
         ]);
 
         $headers = [ 'Content-Type' => 'text/csv; charset=utf-8' ];
 
-        return response()->streamDownload(function () use ($import_id, $results) {
-            $results->generateCSV('php://output', $import_id);
+        return response()->streamDownload(function () use ($import, $results) {
+            $results->generateCSV('php://output', $import);
         }, $filename, $headers);
     }
 }
--- a/app/Http/Controllers/ImportsResultsController.php
+++ b/app/Http/Controllers/ImportsResultsController.php
@@ -12,7 +12,7 @@ class ImportsResultsController extends Controller
         $import = ImportMeta::findOrFail($import_id);
 
         $filename = strtr(config('imports.results.filename_template'), [
-            ':id' => $import_id
+            ':id' => $import_id,
             ':datetime' => now()->format('Y-m-dTH:i:s')
         ]);
 
--- a/app/Http/Controllers/ImportsResultsController.php
+++ b/app/Http/Controllers/ImportsResultsController.php
@@ -15,7 +15,6 @@ class ImportsResultsController extends Controller
         $headers = [ 'Content-Type' => 'text/csv; charset=utf-8' ];
 
         return response()->streamDownload(function () use ($import_id) {
-            // make the file happen somehow
             ImportResults::class->generateCSV('php://output', $import_id);
         }, $filename, $headers);
     }
--- a/app/Http/Controllers/ImportsResultsController.php
+++ b/app/Http/Controllers/ImportsResultsController.php
@@ -9,7 +9,7 @@ class ImportsResultsController extends Controller
     public function showResultsCsv(int $import_id, ImportResults $results)
     {
         $filename = strtr(config('imports.results.filename_template'), [
-            ':datetime' => now()->format('Y-m-d_H-i-s')
+            ':datetime' => now()->format('Y-m-dTH:i:s')
         ]);
 
         $headers = [ 'Content-Type' => 'text/csv; charset=utf-8' ];
--- a/app/Http/Controllers/ImportsResultsController.php
+++ b/app/Http/Controllers/ImportsResultsController.php
@@ -13,7 +13,7 @@ class ImportsResultsController extends Controller
         $headers = [ 'Content-Type' => 'text/csv; charset=utf-8' ];
 
         return response()->streamDownload(function () use ($import_id) {
-            $report->generateCSV('php://output');
+            // make the file happen somehow
         }, $filename, $headers);
     }
 }
--- a/app/Http/Controllers/ImportsResultsController.php
+++ b/app/Http/Controllers/ImportsResultsController.php
@@ -6,7 +6,7 @@ use App\Services\ImportResults;
 
 class ImportsResultsController extends Controller
 {
-    public function showResultsCsv(int $import_id)
+    public function showResultsCsv(int $import_id, ImportResults $results)
     {
         $filename = strtr(config('imports.results.filename_template'), [
             ':datetime' => now()->format('Y-m-d_H-i-s')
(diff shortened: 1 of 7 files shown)


tests/Feature/ImportReportTest.php
```diff
@@ -21,9 +21,7 @@ class ImportReportTest extends TestCase
     public function test_generate_csv_writes_to_path(): void
     {
         $user = User::factory()->uploader()->create();
-        $import = ImportMeta::factory()->for($user)->create(
-            ['status' => 'completed']
-        );
+        $import = ImportMeta::factory()->for($user)->create();
```
lucaswerkmeister, line 24: Is this change needed? The tests still seem to pass without it.

tests/Feature/ImportResultsTest.php
```diff
@@ -0,0 +1,81 @@
...
+            ->for($import)
+            ->create();
+
+        Storage::fake('local');
+        $filename = 'temp_test.csv';
+
+        Storage::put($filename, '');
+        $path = Storage::path($filename);
+
+        $this->travelTo(now()); // stop the clock
```
lucaswerkmeister, line 35: Do we need the time travel? As far as I can tell, nothing in the CSV depends on it, and the test still seems to pass without it.

app/Http/Controllers/ImportsResultsController.php
```diff
@@ -0,0 +1,21 @@
...
+namespace App\Http\Controllers;
+
+use App\Services\ImportResults;
+
+class ImportsResultsController extends Controller
+{
+    public function showResultsCsv(int $import_id)
+    {
+        $filename = strtr(config('imports.results.filename_template'), [
+            ':datetime' => now()->format('Y-m-d_H-i-s')
```
lucaswerkmeister, line 12: I’d use colons instead of dashes in the “time” part, and preferably also a `T` instead of `_` to separate the “date” and “time” parts, to match ISO 8601.

app/Services/ImportResults.php
```diff
@@ -0,0 +1,43 @@
...
+                $mismatch['wikidata_value'],
+                $mismatch['meta_wikidata_value'],
+                $mismatch['external_value'],
+                $mismatch['external_url'],
+                $mismatch['review_status']
+            ];
+        }, $mismatches->toArray());
+
+        // Add column keys as CSV headers
+        $mismatchesWithKeys = array_merge([config('imports.results.column_keys')
```
lucaswerkmeister, line 34: IMHO it would be nicer to emit the headers in a separate `fputcsv()` call (like in `ImportReport`), rather than merge two fairly different arrays together.

app/Services/ImportResults.php
```diff
@@ -0,0 +1,43 @@
...
+    {
+        $mismatches = Mismatch::get();
+
+        // Creating the output stream
+        $handle = fopen($path, 'w');
+
+        $import = ImportMeta::find($import_id);
+        $mismatches = Mismatch::whereBelongsTo($import)->get();
+
+        $mismatchValues = array_map(function ($mismatch) {
```
lucaswerkmeister, line 20: I think instead of  mapping the whole result at once, and then iterating over the resulting array, we could also iterate over the mismatches directly and call `fputcsv()` for each row?

lucaswerkmeister:
By the way, this pull request also already implements T325177.

tests/Feature/ImportReportTest.php
chukarave: It was left over from the test I took the code from but was redundant for this test therefore removed

chukarave: yeah this seems to be unintentional

tests/Feature/ImportResultsTest.php
chukarave: removed

app/Services/ImportResults.php
chukarave: I ended up making this a LazyCollection, where the mapping is only triggered by the `.each()` iteration which writes the mismatch to the csv

tests/Feature/ImportReportTest.php
```diff
@@ -46,8 +46,6 @@ public function test_generate_csv_writes_to_path(): void
         $total = $mismatches->count();
         $pending = $mismatches->where('review_status', 'pending')->count();
 
-        $this->travelTo(now()); // stop the clock
```
itamargiv, line 49: This change doesn't necessarily relate to the ticket we are working, I think it will probably be best to do this in a separate PR.

config/imports.php
```diff
@@ -51,5 +51,19 @@
             "expiry date",
             "expired?"
         ]
+    ],
+
+    'results' => [
+        'filename_template' => 'import-results-:datetime.csv',
```
itamargiv, line 57:
```suggestion
        'filename_template' => 'results-:id-:datetime.csv',
```

According to T325177 the filename should be a bit more explicit, like so.

app/Http/Controllers/ImportsResultsController.php
```diff
@@ -0,0 +1,24 @@
...
+
+class ImportsResultsController extends Controller
+{
+    public function showResultsCsv(int $import_id, ImportResults $results)
+    {
+        $import = ImportMeta::findOrFail($import_id);
+
+        $filename = strtr(config('imports.results.filename_template'), [
+            ':datetime' => now()->format('Y-m-dTH:i:s')
+        ]);
```
itamargiv, line 16:
```suggestion
            ':id' => $import_id
            ':datetime' => now()->format('Y-m-dTH:i:s')
        ]);
```

In accordance with the file name template change suggested in the comment above.

app/Services/ImportResults.php
```diff
@@ -0,0 +1,40 @@
...
+            return [
+                $mismatch['item_id'],
+                $mismatch['statement_guid'],
+                $mismatch['property_id'],
+                $mismatch['wikidata_value'],
+                $mismatch['meta_wikidata_value'],
+                $mismatch['external_value'],
+                $mismatch['external_url'],
+                $mismatch['review_status']
+            ];
```
itamargiv, line 33: In a situation where we need to add a column, hard-coding the column keys here as well will require us to change two separate files.

However, if we map each key (as we get it from the config, into it's value here, we can avoid this redundancy)

app/Services/ImportResults.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    public function generateCSV(string $path, ImportMeta $import): void
+    {
+        $mismatches = Mismatch::get();
+
+        // Creating the output stream
+        $handle = fopen($path, 'w');
+
+        // Get the mismatches as a LazyCollection
+        $mismatches = Mismatch::whereBelongsTo($import)->cursor();
```
itamargiv, line 18: Thanks for using a lazy collection, it's a good idea. Also, the comment makes it explicit so that we know which docs to refer to.

tests/Feature/ImportResultsTest.php
```diff
@@ -0,0 +1,77 @@
...
+            return [
+                $mismatch['item_id'],
+                $mismatch['statement_guid'],
+                $mismatch['property_id'],
+                $mismatch['wikidata_value'],
+                $mismatch['meta_wikidata_value'],
+                $mismatch['external_value'],
+                $mismatch['external_url'],
+                $mismatch['review_status']
+            ];
```
itamargiv, line 45: Same as the prior comment, we can reduce redundancy here as well, so that we don't have to keep this updated in several files.

tests/Feature/WebStoreRouteTest.php
```diff
@@ -64,4 +66,27 @@ public function test_store_import_csv_stats()
...
+    {
+        $response = $this->get(route('store.import-results', ['123123123']));
+        $response->assertStatus(404);
+    }
+
+    public function test_store_import_results_route()
+    {
+        $filename = strtr(config('imports.results.filename_template'), [
+            ':datetime' => now()->format('Y-m-dTH:i:s')
+        ]);
```
itamargiv, line 80:
```suggestion
            ':id' => $import->id
            ':datetime' => now()->format('Y-m-dTH:i:s')
        ]);
```

Same as the above comments regarding the change in the filename template.
ofc this will also require us to determine the ImportMeta prior to the filename assignment.

itamargiv (changes requested):
Looks great! I've left a couple of comments, but besides that everything works great while running locally.

itamargiv (approved):
Great work! Thank you!

PR #514, opened 2022-12-30
Remove redundant time traveling

chukarave:
The test does not require any time manipulation anymore therefore these may be removed.


guergana (approved):
Indeed seems to not be needed. Thanks for removing this unneeded code.

PR #525, opened 2023-01-02
Add 'download csv' button to imports

guergana:
Bug: T304794


public/mix-manifest.json
```diff
@@ -1,7 +1,10 @@
 {
     "/js/app.js": "/js/app.js",
     "/css/app.css": "/css/app.css",
+    "/images/download-stats-black.svg": "/images/download-stats-black.svg",
+    "/images/download-stats-white.svg": "/images/download-stats-white.svg",
     "/images/download-stats.svg": "/images/download-stats.svg",
```
micgro42, line 6: question: As you renamed one usage of the `download-stats.svg` file to `download-stats-white.svg`, is this line here still needed?

public/mix-manifest.json
```diff
@@ -1,7 +1,10 @@
...
     "/images/mismatch-finder-logo.svg": "/images/mismatch-finder-logo.svg",
-    "/images/user.svg": "/images/user.svg"
+    "/images/user.svg": "/images/user.svg",
+    "/images/wikidata-logo.svg": "/images/wikidata-logo.svg"
```
micgro42, line 9: suggestion: This line about the logo seems unrelated to this change.

I assume it was added by mistake and should probably be dropped?

resources/sass/app.scss
```diff
@@ -250,6 +250,10 @@ dl.import-meta dd:not(:last-child) {
     padding-bottom: 16px;
 }
 
+dl.import-meta .download-csv {
+    float:right;
```
micgro42, line 254: nitpick (non-blocking): missing space after the colon

(Though we really ought to have linting for that)

resources/views/importStatus.blade.php
```diff
@@ -1,9 +1,10 @@
 <x-layout>
     <div role="group" class="button-group right">
-        <a class="button primary progressive" href="{{ route('store.imports-overview') }}"><img src="{{ asset('/images/download-stats.svg') }}" class="button-icon" />{{__('store-layout.button:download-stats')}}</a>
+        <a class="button primary progressive" href="{{ route('store.imports-overview') }}"><img src="{{ asset('/images/download-stats-white.svg') }}" class="button-icon button-white-fill" />{{__('store-layout.button:download-stats')}}</a>
```
micgro42, line 3: Suggestion:  That extra class is probably not needed and can be dropped.

As I understand it, you can only affect SVGs with external CSS if they are literally in the DOM as HTML, not when they are used as an image in an `` tag.

guergana: hi, yes, i mentioned that i am still working on this. ;)

resources/sass/app.scss
guergana: fixed!

public/mix-manifest.json
guergana: oh, yes, i don't know where this is coming from, since this file is automatically generated.

guergana: removed!

micgro42:
Looks good to me purely from inspecting the code.
As this does not have browser tests, we should also make sure we try out the functionality manually. I currently don't have a working setup locally, but maybe @chukarave could give it a quick spin, given that you also have done the work that created these routes and are familiar with all the edge-cases that might exist there :)

resources/lang/en/store-layout.php
```diff
@@ -17,6 +17,7 @@
     'button:delete' => 'Delete',
     'button:confirm' => 'Confirm',
     'button:create' => 'Create',
+    'button:download-import-csv' => 'Download csv',
```
chukarave, line 20: I know this is how it was written in Figma, but as it is an acronym i think we want to write it in upper case.

```suggestion
    'button:download-import-csv' => 'Download CSV',
```

chukarave:
It looks good and works well! 
I wonder if we should a dd a media query to remove the float on mobile, as the button width squashes the `Import Status` label (This is Firefox, it looks better on Chrome but i still think it might make sense). Tagging @sai-san to have a look.
(with float)
image

(without)
image

resources/lang/en/store-layout.php
guergana: Oh, i just saw the suggestion, I have pushed the change already. Thanks!!

sai-san:
> I wonder if we should add a media query to remove the float on mobile, as the button width squashes the `Import Status` label (This is Firefox, it looks better on Chrome but i still think it might make sense).
> (without) image

Apologies for the lack of responsive specifications here. I agree with  @chukarave's suggestion and think that the button should simultaneously become full-width and be placed on top of the Import Status card at the 720px media query, @width-breakpoint-tablet (the same that's altering the size and placement of the 'Download statistics button'). 
When placed at the top, the 'Download CSV' buttons should display a `margin-block-end` of 1em that separates them from the first <dt> ('Import status').

chukarave (approved):
Looks great, thank you!

PR #527, opened 2023-01-04
Upgrade version of @typescript-eslint/eslint-plugin

guergana:
Bug: [T326196](https://phabricator.wikimedia.org/T326196)


package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "wikidata-mismatch-finder",
+    "name": "html",
```
lucaswerkmeister, line 2: This doesn’t look nice – we should probably add the `name` to `package.json` instead. (Looks like @chukarave added it to `package-lock.json` in #489?)

lucaswerkmeister: ⇒ #528

guergana: This is automatic. Yeah, i guess it will be better in package.json

lucaswerkmeister: I merged that PR, can you try rebasing / repeating this PR and see what happens to `package-lock.json` now?

lucaswerkmeister (approved):

PR #528, opened 2023-01-04
Add name to package.json

lucaswerkmeister:
Previously, it was only added to package-lock.json, where it’s prone to being removed again by other package operations.


guergana (approved):
ok! :D

PR #532, opened 2023-01-12
Further fixes for CSV download button styles

lucaswerkmeister:
See individual commit messages.

Bug: T304794


sai-san (approved):
Looks perfect! The buttons are now consistent with WiKit 💯

PR #548, opened 2023-02-02
Fix tablet breakpoint inconsistency

lucaswerkmeister:
We were using the tablet breakpoint as max-width for the logo, but as min-width for the header flex-direction, resulting in a one-pixel discrepancy. Let’s use min-width consistently.

Bug: T324357


guergana (approved):
Looks great, thanks..

PR #578, opened 2023-04-20
Localisation updates from https://translatewiki.net.

translatewiki:
Translation updates


itamargiv (approved):
Validated Locally, and it works

PR #600, opened 2023-05-19
Add instructions for updating i18n files

itamargiv:


docs/README.md
```diff
@@ -397,6 +397,15 @@ If, after following the steps in the [Working with OAuth section](#oauth), you a
 1. Make sure that your `APP_URL` is set correctly to your localhost address and the port your application is running on (if running on a non default port).
 1. Double check to see that the WMF has emailed to approve your localhost OAuth consumer, if they have not responded yet, please wait patiently, as the review process might take a few hours.
 
+## Chore: Updating i18n files
+
+Periodically, translatewiki.net will open a PR against our repository with translation updates. In order to make sure our interface is fully localized, we should review and merge these changes:
+
+1. Check out the branch for the i18n update PR.
+1. Run mismatch finder locally with `sail up`
+1. Open the your local envoronment in the browser and add the `uselang` parameter with a language code that is included in the change (for example `uselang=he`, if the change includes a `he.json` file).
```
mariushoch, line 406: Do we really need to do that? Seems quite time consuming and in other projects we just trust the translations (although there can be occasional issues with RTL languages and too large to fit labels, but catching all of them would require going through the whole application).

lucaswerkmeister: Yeah, I’m also not a fan of manually reviewing i18n changes. But if there are things we can test automatically, we should do that (e.g. ensure that translations contain no tags other than `<a>`, and no attributes other than `href=`, judging by the current `en.json` contents?), so that those tests run automatically in CI on those pull requests.

lydiapintscher: FWIW I really think we're doing our users a disservice by letting this sit for 3 weeks+.

mariushoch: > Yeah, I’m also not a fan of manually reviewing i18n changes. But if there are things we can test automatically, we should do that (e.g. ensure that translations contain no tags other than `<a>`, and no attributes other than `href=`, judging by the current `en.json` contents?), so that those tests run automatically in CI on those pull requests.

While this sounds like a good idea, I would deem it optional for now (as a quick glance over the translation changes would also catch trivial HTML injection, like script tags).
Ping @itamargiv: What do you think?

itamargiv:
I am closing this PR as this is no longer relevant after the team decided on a new branching model which does not involve automatic deployments to production. Therefor, we can allow direct commits into the new default branch: `development`.

(See: https://phabricator.wikimedia.org/T332695#9150273)

PR #601, opened 2023-05-21
Fix "ticket" to "task" in a message

amire80:
The usual Phabricator terminology is "task",
not "ticket".


itamargiv (approved):
Thank you!

PR #602, opened 2023-05-22
Localisation updates from https://translatewiki.net.

translatewiki:
Translation updates


lucaswerkmeister (approved):
No bad HTML in the translations AFAICT, just `<a href=""></a>` (and, once in `ms.json`, `class="notranslate" translate="no">`).

PR #617, opened 2023-06-06
Fix Sail dependency version

lucaswerkmeister:
The dependency on v1.51.1 was added in commit 6c7b77f41e (#380), but that version doesn’t actually exist (the latest release is v1.22.0); since the latest version at the time was v1.15.1, I assume the two digits were just swapped accidentally (and apparently composer didn’t complain).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "facade/ignition": "^2.17.7",
         "fakerphp/faker": "^1.20.0",
         "laravel/dusk": "^6.25.1",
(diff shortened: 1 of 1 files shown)


mariushoch:
>[…] (and apparently composer didn’t complain).

Which is _really_ odd… this definitely shouldn't have worked. Apparently our CI is not installing the dev dependencies, so chances are that this never worked :/ 
Maybe we should have CI test that? But then again, I'm not sure how much we should care.

PR #619, opened 2023-06-07
composer update

lucaswerkmeister:
Bug: T338222


lucaswerkmeister:
(This reduces the `composer outdated` output by a lot, but there are still some outdated packages left that will be upgraded separately.)

mariushoch (approved):

PR #639, opened 2023-07-27
Use the latest version of chrome-driver instead of detecting OS' version

hasanakg:
According to latest changes on coverage of chromium.org (hence dusk:chrome-driver uses it for downloads and queries), they no longer support chrome versions newer than v114 (https://chromedriver.chromium.org/downloads), to avoid any inconvenience until dusk:chrome-driver is updated, we should use the latest version.


itamargiv (approved):
Thank you, I will create a task to update dusk

PR #657, opened 2023-08-29
Install latest chrome version for testing

hasanakg:
Bug: T345094

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -82,7 +82,12 @@ jobs:
           php artisan key:generate
       - name: Upgrade Chrome Driver
         run: php artisan dusk:chrome-driver
-
+      - name: Install Chrome
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -82,7 +82,12 @@ jobs:
           php artisan key:generate
       - name: Upgrade Chrome Driver
         run: php artisan dusk:chrome-driver
-
+      - name: Install Chrome
+        run: |
```
guergana, line 86: amazing, thanks. :)

guergana (dismissed):
Looks great. I approve but dunno if @itamargiv has any remarks?

itamargiv (changes requested):
I mean, it looks good to me, but we need to remember to remove this when we get to update dusk, since the newer version of dusk doesn't have this issue IIRC (It was some problem with the request url construction on the chrome driver download in dusk, right?)

Regardless, I think we should add a comment to remind us what this does and to remove it in case we update dusk. Perhaps even an associated task would be good here.

PR #675, opened 2023-09-07
Upgrade composer dependencies

guergana:
Bug: T345615


itamargiv (approved):
Looks good to me! Tests seem to pass and no issues on local run

PR #678, opened 2023-09-18
Add db 'type' column to migration

guergana:
The seemingly unrelated CSV validation tests have been updated
as well to test the change is working. To test the changes run
`sail artisan migrate:refresh --seed`

Side note: As a (hacky) workaround for the migration errors with the enum 
type when running `php artisan migrate` or rollbacks see the commit 
message in [PR 474](https://github.com/wmde/wikidata-mismatch-finder/pull/474)
the `strict` config in `config/database` can be temporarily set to `false`. 

Bug: [T313467](https://phabricator.wikimedia.org/T313467)

### DIFF
--- a/app/Models/Mismatch.php
+++ b/app/Models/Mismatch.php
@@ -22,7 +22,8 @@ class Mismatch extends Model
         'meta_wikidata_value',
         'external_value',
         'external_url',
-        'review_status'
+        'review_status',
+        'type'
     ];
 
     /**
@@ -31,7 +32,8 @@ class Mismatch extends Model
      * @var array
      */
     protected $attributes = [
-        'review_status' => 'pending'
+        'review_status' => 'pending',
+        'type' => 'statement'
     ];
 
     public function importMeta()
--- a/config/imports.php
+++ b/config/imports.php
@@ -15,7 +15,8 @@ return [
             'wikidata_value',
             'meta_wikidata_value',
             'external_value',
-            'external_url'
+            'external_url',
+            'type'
         ]
     ],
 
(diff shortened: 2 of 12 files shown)


itamargiv (changes requested):
Thank you for all the work! I haven't reviewed everything just yet, but I did already notice something that requires a change.

That is, I don't think we should change all the past migrations. The idea behind a migration is to keep a track record of all changes we did to our database, and be able to roll back if needed. In our case, what I think we should do is add a new migration file to add the type column to the mismatches table in our db (and potentially populate it with "statement", for all existing mismatches). 

You can see:
https://laravel.com/docs/8.x/migrations
and particularly
https://laravel.com/docs/8.x/migrations#columns

guergana:
> Thank you for all the work! I haven't reviewed everything just yet, but I did already notice something that requires a change.
> 
> That is, I don't think we should change all the past migrations. The idea behind a migration is to keep a track record of all changes we did to our database, and be able to roll back if needed. In our case, what I think we should do is add a new migration file to add the type column to the mismatches table in our db (and potentially populate it with "statement", for all existing mismatches).
> 
> You can see: https://laravel.com/docs/8.x/migrations and particularly https://laravel.com/docs/8.x/migrations#columns

Thanks for the great observation @itamargiv , you are absolutely right. I have modified the files :)

itamargiv (approved):
Looks good, very straightforward. I validated it locally by running the migration and inspecting the DB. I also ran the seeder and everything seems in order. 

Please keep in mind, that after merging this we will have to migrate and seed the staging environment as described in https://wikitech.wikimedia.org/wiki/Tool:Wikidata_Mismatch_Finder#Running_Database_Migrations (make sure to `become mismatch-finder-staging` instead of `mismatch-finder`)

PR #681, opened 2023-09-25
Add Download CSV button only on completed imports

guergana:
Bug: [T335864](https://phabricator.wikimedia.org/T335864)


itamargiv (approved):
Looks great, I love one liners :D. Good job!

PR #682, opened 2023-09-25
Make footer stay at the bottom of the page aka sticky

guergana:
Bug: [T344336](https://phabricator.wikimedia.org/T344336)

### DIFF
--- a/resources/js/Components/WikidataToolFooter.vue
+++ b/resources/js/Components/WikidataToolFooter.vue
@@ -72,8 +72,6 @@ export default Vue.extend({
 
 .footer-container {
     background-color: $color-base-90;
-    flex-grow: 0;
-    flex-shrink: 0;
 
     & > footer {
         margin: auto;
--- a/resources/js/Components/WikidataToolFooter.vue
+++ b/resources/js/Components/WikidataToolFooter.vue
@@ -72,6 +72,8 @@ export default Vue.extend({
 
 .footer-container {
     background-color: $color-base-90;
+    flex-grow: 0;
+    flex-shrink: 0;
 
     & > footer {
         margin: auto;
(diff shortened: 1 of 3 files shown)


resources/js/Components/WikidataToolFooter.vue
```diff
@@ -72,6 +72,8 @@ export default Vue.extend({
 
 .footer-container {
     background-color: $color-base-90;
+    flex-grow: 0;
+    flex-shrink: 0;
```
itamargiv, line 76: I think these rules are actually not needed in our case, since the footer sits outside of main in the html of the `Layout.vue` component. When I disable them in the devtools, it has no apparent affect.

resources/sass/app.scss
```diff
@@ -5,8 +5,20 @@
...
 body {
     @include body-text;
+    height: 100%;
+}
+
+#app {
+    box-sizing: border-box;
+    min-height: 100%;
+    display: flex;
+    flex-direction: column;
```
itamargiv, line 21: That’s great and works fantastically. Though, since these rules are more thematically related to the layout, I'd say we should move them into `Layout.vue` under the `.website` selector (it's actually the same element as `#app` after Vue is initialized).

itamargiv (changes requested):
This looks great, thank you! I left a couple of comments after validating locally.

resources/sass/app.scss
itamargiv: Ah no, I'm suggesting you move these rules there since they mostly relate to the html that is being produced by `Layout.vue`. In this file I would only add rules that relate to `html` and `body` or very general rules that are agnostic to the content of the app (such as reset rules for example). Although I do see other rules in this sass file that I think should move to more appropriate places, but that's for another task, not this one.

itamargiv (approved):
Great work, thank you!

PR #686, opened 2023-09-28
Add custom noscript page for Javascript disabled browsers

hasanakg:
Bug: T343344

### DIFF
--- a/resources/sass/noscript.scss
+++ b/resources/sass/noscript.scss
@@ -1,6 +1,6 @@
 @import '~@wikimedia/codex/dist/codex.style.css';
 @import '~@wikimedia/codex/dist/codex.style-rtl.css';
-@import '~@wikimedia/codex-design-tokens/dist/theme-wikimedia-ui.scss';
+@import '~@wikimedia/codex-design-tokens/theme-wikimedia-ui.scss';
 
 $spacing-layout-50: 0.5rem;
 $spacing-layout-100: 1rem;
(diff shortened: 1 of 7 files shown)


resources/sass/noscript.scss
```diff
@@ -0,0 +1,65 @@
+@import '~@wikimedia/codex/dist/codex.style.css';
```
chukarave, line 1: Do you know why we have to import from `/dist/`? I saw we do it for the app css too but it's not how the [codex docs](https://doc.wikimedia.org/codex/main/using-codex/usage.html#using-design-tokens) advise to do it

resources/views/app.blade.php
```diff
@@ -5,6 +5,29 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0" />
     <link href="{{ mix('/css/app.css') }}" rel="stylesheet" />
     <script src="{{ mix('/js/app.js') }}" defer></script>
+
+    <noscript>
```
chukarave, line 9: My personal preference would be to include this in a separate template that gets included here. less clutter

chukarave (commented):
It works well for me locally, as far as I can tell the styles in the breakpoints are correct. some very minor things.

resources/views/app.blade.php
hasanakg: You mean having a separate `noscript.blade.php` file right? We already have a separate `scss` file for that, so it sounds better for me as well. If that's the case, just let me know so I can update the code.

chukarave: yes that's exactly what i mean!

chukarave (approved):
The latest version looks good to me.

PR #687, opened 2023-09-28
Add type column gui

guergana:
Bug: [T313467](phabricator.wikimedia.org/T313467)


public/i18n/qqq.json
```diff
@@ -25,7 +25,8 @@
 	"item-form-error-message-max": "An error message when too many item ids are provided",
 	"item-form-error-message-invalid": "An error message when invalid Item ids are provided",
 	"server-error": "An error message when the server encountered an error retrieving results",
-	"column-property": "Property column on mismatches table",
+	"column-mismatch": "Property column on mismatches table",
```
chukarave, line 28: I guess we should change this to say something else? so the people translating will know what this message is about. Though something like "mismatch column on mismatches table" sounds a bit meaningless to me, maybe you have a better idea?

chukarave: I thought this was being changed from "property" to "mismatch" to allow adding them for qualifiers, so leaving the word property there has potential for confusion, imo. Though I agree "Mismatch column etc." is no good.

guergana: I guess we have to reach an agreement on what is good. Mismatch column could be ok but sounds a bit repetitive

chukarave (approved):
looks alright to me, thanks

amire80:
Beep, an important comment about the translation files.

Before I go on: This particular issue doesn't require any immediate action because I've already fixed it on translatewiki by deleting the incorrect translations.

However, for the future, please make sure that you change keys in the JSON translation files differently. In this case, you should have done the following:
* Delete the unnecessary key from en.json and qqq.json.
* Add the new key to en.json and qqq.json.
* **Don't do anything with the other languages.** The removed key would be automatically deleted by a bot in the next sync.

What happened is that you renamed the keys in all the languages, but left the old translation, so "Mismatch" became quietly translated as "Property" to all of them.

In general, touching the translation files (other than en.json and qqq.json) is usually not necessary at all, and they are automatically updated by the translatewiki bot. If you think you have a good reason to change something in them, I recommend consulting with translatewiki maintainers at https://translatewiki.net/wiki/Support or at the Telegram group (the link appears at the top of the Support page).

Thanks :)

guergana:
> Beep, an important comment about the translation files.
> 
> Before I go on: This particular issue doesn't require any immediate action because I've already fixed it on translatewiki by deleting the incorrect translations.
> 
> However, for the future, please make sure that you change keys in the JSON translation files differently. In this case, you should have done the following:
> 
>     * Delete the unnecessary key from en.json and qqq.json.
> 
>     * Add the new key to en.json and qqq.json.
> 
>     * **Don't do anything with the other languages.** The removed key would be automatically deleted by a bot in the next sync.
> 
> 
> What happened is that you renamed the keys in all the languages, but left the old translation, so "Mismatch" became quietly translated as "Property" to all of them.
> 
> In general, touching the translation files (other than en.json and qqq.json) is usually not necessary at all, and they are automatically updated by the translatewiki bot. If you think you have a good reason to change something in them, I recommend consulting with translatewiki maintainers at https://translatewiki.net/wiki/Support or at the Telegram group (the link appears at the top of the Support page).
> 
> Thanks :)

Hello @amire80 Thanks for the explanation. I didn't touch the translation files in any way or added any keys. In fact they appeared on their own in the PR. I decided to keep them there since I am not familiar with translatewiki and assumed this is something that was done automatically. Maybe this happened when rebasing, but I didn't manually edit any of these files. If this happens again should I drop them from my commit?

amire80:
Yes, please. There is usually no reason to do anything manually in the json files with translations, because they are managed by the translatewiki bot. Only en.json and qqq.json need to be edited manually.

guergana:
> Yes, please. There is usually no reason to do anything manually in the json files with translations, because they are managed by the translatewiki bot. Only en.json and qqq.json need to be edited manually.

Oh, ok. Will keep this in mind in case some translations pop up again when rebasing. Thanks for the heads up. :) @amire80

PR #751, opened 2023-11-03
Install vue3 and dependencies

hasanakg:


itamargiv (approved):
Reviewed during mob review session with @hasanakg @chukarave .

PR #752, opened 2023-11-03
Setup pinia

hasanakg:


itamargiv (approved):
Reviewed in a mob review session with @hasanakg and @chukarave

PR #753, opened 2023-11-03
Setup i18n

hasanakg:


itamargiv (approved):
Reviewed during mob review session with @hasanakg and @chukarave

PR #755, opened 2023-11-03
Replace wikit dialog with codex dialog

hasanakg:


itamargiv (approved):
Reviewed in mob review session with @hasanakg and @chukarave

PR #756, opened 2023-11-03
Fix deprecation errors

hasanakg:


itamargiv (approved):
Reviewed together with @hasanakg

PR #764, opened 2023-11-13
Replace wikit button with codex button

guergana:
Bug: [T346797](https://phabricator.wikimedia.org/T346797)

### DIFF
--- a/resources/js/Pages/Home.vue
+++ b/resources/js/Pages/Home.vue
@@ -4,17 +4,15 @@
         <section id="description-section">
             <header class="description-header">
                 <h2 class="h4">{{ $i18n('about-mismatch-finder-title') }}</h2>
-                <wikit-button
+                <cdx-button
                     id="faq-button"
-                    variant="quiet"
-                    type="progressive"
-                    @click.native="faqDialog = true"
+                    weight="quiet"
+                    action="progressive"
+                    @click="faqDialog = true"
                 >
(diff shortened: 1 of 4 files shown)


resources/js/Pages/Home.vue
```diff
@@ -226,6 +221,12 @@
                 validationError: null,
                 faqDialog: false
             }
+        },
+        setup() {
```
chukarave, line 225: Do you think this should be higher up in the object order? Below components for example? Just for visibility

resources/js/Pages/Layout.vue
```diff
@@ -178,6 +177,11 @@ export default defineComponent({
     beforeUnmount () {
         this.resizeObserver.unobserve(this.$refs.contentWrap as Element)
     },
+    setup() {
```
chukarave, line 180: Same question as above, should this be moved higher up for visibility?

chukarave (commented):
Two nit-picky things, functionality works well :+1: 

Considering Sarai's unavailability I think we can merge this as it is and iterate on the styles once she's back.

resources/js/Pages/Home.vue
guergana: Yes, having setup at the very beginning sounds like a good idea. i thought the linter would complain and save me from deciding on my own.

chukarave (approved):
codex button galore

PR #766, opened 2023-11-16
Update deps

chukarave:
@typescript-eslint
vue-loader
all codex packages

Bug: T350667


guergana (commented):
I see the tests are passing, but I see a major version change in codex. Are we sure the element(s) that we are using from codex now are still working as expected?

chukarave:
As far as i've tested they do but would appreciate more eyes testing!

guergana (approved):
LGTM!

hasanakg (approved):

PR #767, opened 2023-11-16
update chore docs after Vue3 upgrade

chukarave:
Now that upgraded the mismatch finder to use Vue 3 and are gradually deprecating Wikit, Vue packages should be upgraded normally. 

Bug:  T350667


guergana (approved):
Good catch! LGTM!

PR #771, opened 2023-11-20
Replace Wikit Textarea with Codex Textarea component

guergana:
This PR creates a wrapper around the Textarea component in Vue3.

Bug: T347190


resources/js/Pages/Home.vue
```diff
@@ -132,24 +135,40 @@
...
+            const textareaInputValue = ref(store.lastSearchedIds);
+            
             return {
                 cdxIconDie,
-                cdxIconInfo
+                cdxIconInfo,
+                textareaInputValue,
+                form: {
+                    itemsInput: store.lastSearchedIds
+                },
```
itamargiv, line 166: Wait, so do we even still need this now that we have textAreaInputValue?

itamargiv (commented):
Looks good so far, left my bits and bobs. But I would still encourage @hasanakg and @chukarave to leave their review.

resources/js/Pages/Home.vue
guergana: Ah, no, it is not needed. I thought i had removed it.

guergana: > Wait, so do we even still need this now that we have textAreaInputValue?

Ah, no, it's not needed. I thought I had removed it already.

resources/js/Pages/Home.vue
```diff
@@ -162,20 +181,22 @@
             validate(): void {
                 this.validationError = null;
 
+                const typeError = 'error';
+
                 const rules = [{
                     check: (ids: Array<string>) => ids.length < 1,
-                    type: 'warning',
-                    message: this.$i18n('item-form-error-message-empty')
+                    type: typeError,
```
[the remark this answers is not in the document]
guergana: I added a variable because at some point there will be a warning and because it needs to be the same name in the type and they key of the message. So they are not changed separately.

resources/js/Pages/Home.vue
```diff
@@ -132,24 +135,40 @@
...
           CdxMessage,
-          InertiaHead,
-          TextArea,
+          CdxProgressBar,
+          CdxTextArea,
+          InertiaHead
         },
         setup() {
+            const store = useStore();
+            const textareaInputValue = ref(store.lastSearchedIds);
```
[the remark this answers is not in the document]
guergana: Yes, Pinia looks a lot more compact :)

chukarave:
This looks good to me and works well locally. It's a :+1: from me but i'll let @itamargiv approve since a review has already started.

hasanakg (commented):
Overall seems fine but why didn't we create a new wrapper component instead of crafting it inside `Home.vue` file?

hasanakg:
> > Overall seems fine but why didn't we create a new wrapper component instead of crafting it inside `Home.vue` file?
> 
> Hi. I didn't think it was necessary. The new component doesn't add a lot of complexity compared to the other one, except for the fact that it has some nested elements.

You are right about complexity, but shouldn't it be a reusable component? I checked our plan of action on the ticket and our plan was also wrapping it as a component. T347190

resources/js/Pages/Home.vue
```diff
@@ -138,64 +131,31 @@
...
-                if(this.validationError) {
+                if((this.$refs.textarea as InstanceType<typeof TextareaHome>).validationError) {
                     return;
                 }
                 const store = useStore();
-                store.saveSearchedIds( this.form.itemsInput );
-                this.$inertia.get( '/results', { ids: this.serializeInput() } );
+                store.saveSearchedIds( this.textareaInputValue );
+                this.$inertia.get( '/results', 
+                    { ids: (this.$refs.textarea as InstanceType<typeof TextareaHome>).serializeInput() }
```
itamargiv, line 157: I'd prefer it if we didn't directly call component methods. While this can be done with composition API using [`defineExpose`](https://vuejs.org/api/sfc-script-setup.html#defineexpose), the best practice here would be to emit an event with the already serialized input and use that event to save them to the component state. Then, this `send` method becomes much simpler, as it only needs to read the IDs from the state.

itamargiv (changes requested):
So far so good, I have left some comments even before testing locally. I will look into it later on to see why the tests are failing.

itamargiv:
Alright, so I was looking into the testing errors and the problem is as always... Typescript. It actually doesn't build locally for me without raising errors. I think the issue is that we are using the `this` keyword on the `$i18n` calls within the methods, for example, in [`TextareaHome.vue:72`](https://github.com/wmde/wikidata-mismatch-finder/blob/6028a8f19053e5512ac6b5b0f6d12a0ac3f043a9/resources/js/Components/TextareaHome.vue#L72) and the lines below it. Hopefully this should resolve once the component is converted to the composition API (and we will just have to use `$i18n` without the `this` keyword) but if not, we will talk about it again afterwards.

resources/js/Components/TextareaHome.vue
```diff
@@ -0,0 +1,121 @@
...
+          :status="validationError ? validationError.type : 'default'"
+          v-model="textareaInputValue"
+      />
+  </cdx-field>
+</template>
+
+<script lang="ts">
+import { defineComponent, ref } from 'vue';
+import { useStore } from '../store';
+import { MAX_NUM_IDS } from '../Pages/Home.vue';
```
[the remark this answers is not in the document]
guergana: true

resources/js/Components/ItemIdSearchTextarea.vue
```diff
@@ -0,0 +1,109 @@
...
+const store = useStore();
+const textareaInputValue = ref(store.lastSearchedIds);
+
+const MAX_NUM_IDS = inject('MAX_NUM_IDS');
+
+defineProps<{loading: boolean}>();
+
+function splitInput(): Array<string> {
+    return textareaInputValue.value.split( '\n' );
+}
```
chukarave, line 47: I think some spaces are needed in this line and after the next couple of functions. So close together is a bit messy on the eyes.

resources/js/Components/ItemIdSearchTextarea.vue
```diff
@@ -0,0 +1,109 @@
...
+// Run it with compat mode
+// https://v3-migration.vuejs.org/breaking-changes/v-model.html
+CdxTextArea.compatConfig = {
+    ...CdxTextArea.compatConfig,
+    COMPONENT_V_MODEL: false,
+};
+
+const validationError: Ref<ValidationError> = ref(null);
+
+const banana = useI18n();
```
chukarave, line 36: Are we sure this is the variable name we want to use for this? I'll accept it if the answer is yes, just want to make sure we give this proper consideration

chukarave (commented):
Some very small nitpicks that should not delay merging

resources/js/Components/ItemIdSearchTextarea.vue
guergana: the plugin is called `vue-banana-i18n`, so "banana" is not such a strange name for the variable. In the use case we found [in the original library](https://github.com/santhoshtr/vue-banana-i18n/blob/master/src/index.js#L52) they call it `bananai18n`, but it's almost the same.

resources/js/Components/ItemIdSearchTextarea.vue
guergana: I've added some spaces.

resources/js/app.ts
```diff
@@ -18,7 +18,9 @@ import Layout from './Pages/Layout.vue';
         const i18nPlugin = createI18n({
             locale: locale,
             messages: i18nMessages,
-            wikilinks: true
+            wikilinks: true,
+            globalInjection: true,
+            legacy: false
```
itamargiv, line 23: There are actually not from the vue-banana-i18n library, we do not need them anymore.

itamargiv (approved):
This looks good to me, thank you. I'll approve, but I still think you need to address Noa's comment about the naming.

itamargiv (dismissed):
🤦 I forgot that there were changes I requested. Sorry. Other than that, I will track other issues in a tech debt log, since I think we can just move on with this.

resources/js/Components/ItemIdSearchTextarea.vue
```diff
@@ -33,7 +33,7 @@ CdxTextArea.compatConfig = {
 
 const validationError: Ref<ValidationError> = ref(null);
 
-const banana = useI18n();
+const messages = useI18n();
```
chukarave, line 36: nice!

chukarave (approved):
Looks good to me (and it was working well all along) :+1:

chukarave (approved):
Approving again

chukarave:
I'm gonna go ahead and merge this so it gets to staging

PR #772, opened 2023-11-21
Replace WikitLink component with Codex styles

chukarave:
This PR adds a new sass file which was converted from codex's [link style less file](https://github.com/wikimedia/design-codex/blob/main/packages/codex/src/themes/mixins/public/link.less) in order for us to be able to introduce the styles to the mismatch finder. 

Since codex has links as css only, the wikit component was replaced with an HTML element which was then styled using the converted mixin.

Bug: T347161


resources/js/Components/AuthWidget.vue
```diff
@@ -41,7 +39,8 @@ export default defineComponent({
         vertical-align: middle;
     }
 
-    .wikit-Link {
+    .cdx-docs-link {
+        @include cdx-mixin-link;
```
itamargiv, line 43: Since the same is done for every component, why not just create the class rules in `app.scss`, this will lead to less duplication. Also what does the `docs` stand for in the class name?

resources/js/Components/MismatchRow.vue
```diff
@@ -199,13 +194,21 @@ export default defineComponent({
...
 <style lang="scss">
 @import '~@wmde/wikit-tokens/dist/_variables.scss';
+@import '~@wikimedia/codex-design-tokens/theme-wikimedia-ui.scss';
+@import '../../sass/link.scss';
 
-    .wikit-Link.break-line-link {
-      width: 100%;
-    }
-    .wikit-Link__content {
+  .cdx-docs-link {
```
itamargiv, line 200: This overwrites the `.break-line-link` class rules but not all wikit links have that class, I would advise to keep these rules separate from the mixin.

resources/js/Components/AuthWidget.vue
```diff
@@ -1,32 +1,30 @@
...
-    },
-    components: {
-        WikitLink
     }
 });
 </script>
 
 <style lang="scss">
 @import '~@wmde/wikit-tokens/dist/_variables.scss';
+@import '~@wikimedia/codex-design-tokens/theme-wikimedia-ui.scss';
```
itamargiv, line 26: Why are these imported here? Aren't they only needed in `link.scss`?

itamargiv (commented):
Looks good all in all. Thank you! I left a few comments, but I would encourage that also @hasanakg or @guergana leave their review as well.

resources/js/Components/AuthWidget.vue
chukarave: not according to the codex docs -
image but i did try it out locally in one spot it doesn't seem to break so we might consider doing it anyway

resources/js/Components/AuthWidget.vue
chukarave: Thinking some more about this comment I realized that we can definitely just import and call the mixin for all link elements in `app.scss` and ended up removing the relevant css from the components and the `cdx-docs-link` class from the templates. 
That means inspecting the links doesn't show any association to cdx link but everything work well, so i'm not sure that's necessary anyway.

resources/js/Pages/Results.vue
```diff
@@ -73,9 +72,9 @@
                 :id="`item-mismatches-${String(item)}`"
                 :key="idx">
                 <h2 class="h4">
-                    <wikit-link :href="`https://www.wikidata.org/wiki/${String(item)}`" target="_blank">
+                    <a :href="`https://www.wikidata.org/wiki/${String(item)}`" target="_blank">
```
guergana, line 75: Adding an anchor element inside a header collapses the margins, this doesn't happen with the wikit-link element.

See:
with anchor (considerably smaller vertical margin)
image

original (wikit-link)
image

resources/js/Pages/Layout.vue
```diff
@@ -40,24 +40,24 @@
             <section>
                 <h2 class="h5">{{ $i18n('mismatch-finder-footer-more-tools') }}</h2>
                 <p>
-                    <wikit-link href="https://query.wikidata.org/querybuilder/">
+                    <a href="https://query.wikidata.org/querybuilder/">
```
guergana, line 43: The margins are collapsed in the footer too, but I see them collapsed only in the results page. o_O

Development: 
image

This branch:
image

guergana (changes requested):

resources/js/Components/AuthWidget.vue
```diff
@@ -1,26 +1,22 @@
...
-        <WikitLink :href="`https://www.wikidata.org/wiki/User:${user.name}`">
+        <a :href="`https://www.wikidata.org/wiki/User:${user.name}`">
             <img src="images/user.svg" class="icon-user" /><span class="username">{{ user.name }}</span>
-        </WikitLink>
-        <WikitLink href="/auth/logout">{{ $i18n('log-out') }}</WikitLink>
+        </a>
+        <a  href="/auth/logout">{{ $i18n('log-out') }}</a>
     </div>
-    <WikitLink v-else href="/auth/login">{{ $i18n('log-in') }}</WikitLink>
+    <a v-else href="/auth/login">{{ $i18n('log-in') }}</a>
```
guergana, line 8: Now there's a focus border around the link that wasn't there 

image

The focus border is also there around the username and the log out links. 

image

guergana (changes requested):

resources/js/Pages/Results.vue
chukarave: good catch, I accidentally moved some styles that applied on wikit-link to the containing element instead of having them on `<a>`

resources/js/Pages/Layout.vue
chukarave: Yeah the styles were on h2 by mistake. it should look better now

resources/js/Components/AuthWidget.vue
chukarave: This is built into codex link. But since we host the mixin locally we could theoretically get rid of it. What does @sai-san think?

sai-san: This is an accessibility improvement to make clear that links are receiving focus. We shouldn't get rid of this style.
I'm wondering about that apparent extra space around the username link, though. I could take a deeper look at what's causing that in staging.

resources/js/Components/MismatchRow.vue
```diff
@@ -200,12 +197,16 @@ export default defineComponent({
...
+  a {
+    &:__content {
       word-break: break-word;
     }
+
+    &:break-line-link {
+      width: 100%
+    }
+  }
+
```
hasanakg, line 209: When using `wikit-link`, it has its own elements inside (a `span` nested in `a` element), these selectors are for them.  I suppose, for our new case something like below should be enough. 

```sass
a {
  &.break-line-link {}
}
```

Note: This is just for demonstration, I didn't add styling on purpose.

hasanakg (changes requested):

resources/js/Components/AuthWidget.vue
chukarave: @sai-san and I discussed this offline and ended up adjusting the styles for the auth widget to fix the above issue

resources/js/Components/MismatchRow.vue
chukarave: Thanks for the hint! I left only the `break-line-link` class, as the other one is redundant as you commented.

guergana (approved):

resources/js/Components/AuthWidget.vue
guergana: Looks great now. Thx

guergana:
approved from my side but I am not sure if @hasanakg or @itamargiv would like to add something from their previous comments.

hasanakg (approved):

PR #773, opened 2023-11-22
Create Vue 3 version of the custom LanguageSelector component

hasanakg:
Bug: T345915

### DIFF
--- a/resources/js/Components/LanguageSelector.vue
+++ b/resources/js/Components/LanguageSelector.vue
@@ -1,140 +1,143 @@
 <template>
-	<div class="mismatchfinder__language-selector">
-		<div class="languageSelector__mobile-header">
-			<span>{{ $i18n( 'language-selector-mobile-header' ) }}</span>
-			<button @click="onCloseMenu">
-				<img :src="closeUrl" :alt="$i18n( 'language-selector-close-button-label' )">
-			</button>
-		</div>
-		<LanguageSelectorInput
-			ref="input"
-			:value="searchInput"
-			:placeholder="$i18n( 'language-selector-input-placeholder' )"
-			@input="onInput"
-			@clear="onClearInputValue"
-			@tab="onCloseMenu"
-			@arrowDown="onArrowDown"
-			@arrowUp="onArrowUp"
-			@enter="onEnter"
-			@escape="onCloseMenu"
-		/>
-		<LanguageSelectorOptionsMenu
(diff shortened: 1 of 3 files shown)


resources/js/Components/LanguageSelector.vue
```diff
@@ -1,140 +1,143 @@
...
+
+defineExpose({focus})
+
+</script>
 <style lang="scss">
 $tinyViewportWidth: 38em;
 
 .mismatchfinder__language-selector {
-	position: absolute;
-	inset-inline-end: 0;
```
guergana, line 108: why is the tab size different now?

hasanakg: I don't know, we don't have any linting rules regarding these kind of things :(

guergana: > I don't know, we don't have any linting rules regarding these kind of things :(

but if we don't have linting rules, what caused it to change? are you using prettier on your local IDE? I would suggest to try to get rid of the new tabbing for now in the PR because this is probably irrelevant for the task. maybe we should address that in a different task if it's needed :pray:

hasanakg: It's probably because of my IDE's default settings, it wasn't my choice anyways. And you are right, I will get rid of them

guergana: thanks :)

hasanakg: now, it's done :)

resources/js/Components/LanguageSelector.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<div class="mismatchfinder__language-selector">
 		<div class="languageSelector__mobile-header">
-			<span>{{ $i18n( 'language-selector-mobile-header' ) }}</span>
+			<span>{{ $i18n('language-selector-mobile-header') }}</span>
```
guergana, line 4: I know this is not 100% consistent across the project but it seems like here there were spaces around the parenthesis and now in this file it has been changed. We need to add a rule to the linter to make this consistent. It's not that relevant for this particular PR but we need to talk about this :)

hasanakg: 100% agree! :)

guergana (approved):
Great job. I don't know if anyone else would like to leave some comment, but it's approved from my side.

PR #774, opened 2023-11-22
Remove CSS not allowing CdxMessage to wrap in small screens

guergana:
Bug: T347183


sai-san (approved):
The fix makes the success message properly adjust its width in viewports below 640px in Chrome v119, Safari v17 or Firefox v119  🙌  💯

PR #775, opened 2023-11-22
Fix cdx-select width problems below 462px viewport width

hasanakg:
Bug: T347168

### DIFF
--- a/resources/sass/app.scss
+++ b/resources/sass/app.scss
@@ -411,3 +411,13 @@ dl.import-meta .download-csv {
 		line-height: $line-height-medium;
 	}
 }
+
+
+.cdx-select-vue {
+	width: 100%;
+	&__handle {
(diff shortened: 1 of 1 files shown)


guergana:
I think it's best if @sai-san reviews this ticket.

sai-san (dismissed):
LGTM! Will try to break this once the changes are reflected on staging 😈

hasanakg:
> LGTM! Will try to break this once the changes are reflected on staging 😈

It's on staging right now, no mercy from now on 😈

sai-san (approved):
LGTM. You don't have to change your process: I can wait to verify when things are merged into development and available in the staging environment. This is of course unless you have doubts about the requirements or find any blockers during implementation. In that case, do not hesitate to pull me in. Thanks!

PR #776, opened 2023-11-22
Fix cdx-dialog styling and remove old wikit-dialog styles

hasanakg:
Bug: T347100

### DIFF
--- a/resources/sass/app.scss
+++ b/resources/sass/app.scss
@@ -3,6 +3,7 @@
 @import '~ress';
 @import '~wikit-dist/wikit-vue-components-vue3compat.css';
 @import "@wikimedia/codex/dist/codex.style.css";
+@import "@wikimedia/codex-design-tokens/theme-wikimedia-ui";
 @import '~@wmde/wikit-tokens/dist/_variables.scss';
 @import './_typography.scss';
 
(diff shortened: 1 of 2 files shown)


resources/sass/app.scss
```diff
@@ -424,4 +401,13 @@ dl.import-meta .download-csv {
 // is marked as resolved.
 .cdx-button .cdx-icon {
     vertical-align: text-top;
-}
\ No newline at end of file
+}
+
+.cdx-dialog {
+	header {
```
sai-san, line 407: There's a little discrepancy here. Let me know what you think: The other heading elements in the Dialog (and the rest of MSMF) have their own selectors in app.scss and are receiving their styling (including their color) from the WiKit mixins included in this file. My guess is that we'll replace said mixins by the necessary Codex tokens to create consistent, reusable text styles (as part of https://phabricator.wikimedia.org/T348732). If so, should we create a h2 selector in app.scss to capture this style there with the rest? We can leave in `.cdx-dialog__header__title` only whatever extra styling is needed in this context. How does that sound?

hasanakg: Thanks a lot, you are right! We can create the mixin for this header with codex tokens as well and that'd make sense as well.

sai-san: ah wait. I just noticed that the Codex implementation is following the same strategy. The main Dialog heading needs to be a h2, but it displays h4 styling. So I guess this is an exception on itself, right? Nothing to change

sai-san (approved):
LGTUX!

PR #792, opened 2023-12-15
Replace all wikit variables in app.scss and _typography.scss

chukarave:
We should probably not merge this before all styles have been updated (as part of [T352314](https://phabricator.wikimedia.org/T352314) and [T352311](https://phabricator.wikimedia.org/T352311)) or it'll break many of the styles on staging.

Bug: T352304

PR #793, opened 2023-12-18
Improve links

chukarave:
Issues this PR addresses:
1. `:visited` and `:hover` styles removed from auth widget links
2. visible outline removed from logo

The issue of auth widget link base color is already part of PR #792 and is therefore not part of this PR.

Bug: T347161

### DIFF
--- a/resources/js/Components/AuthWidget.vue
+++ b/resources/js/Components/AuthWidget.vue
@@ -6,7 +6,7 @@
             </a>
             <a  href="/auth/logout">{{ $i18n('log-out') }}</a>
         </div>
-        <a class="auth-widget" v-else href="/auth/login">{{ $i18n('log-in') }}</a>
+        <a v-else href="/auth/login">{{ $i18n('log-in') }}</a>
     </div>
 
 </template>
--- a/resources/js/Components/AuthWidget.vue
+++ b/resources/js/Components/AuthWidget.vue
@@ -1,10 +1,10 @@
 <template>
     <div class="auth-widget_wrapper"> 
         <div class="auth-widget" v-if="user">
-        <a :href="`https://www.wikidata.org/wiki/User:${user.name}`">
-            <img src="images/user.svg" class="icon-user" /><span class="username">{{ user.name }}</span>
-        </a>
-        <a  href="/auth/logout">{{ $i18n('log-out') }}</a>
+            <a :href="`https://www.wikidata.org/wiki/User:${user.name}`">
+                <img src="images/user.svg" class="icon-user" /><span class="username">{{ user.name }}</span>
+            </a>
+            <a  href="/auth/logout">{{ $i18n('log-out') }}</a>
         </div>
         <a class="auth-widget" v-else href="/auth/login">{{ $i18n('log-in') }}</a>
     </div>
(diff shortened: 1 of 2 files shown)


resources/js/Components/AuthWidget.vue
```diff
@@ -1,11 +1,13 @@
 <template>
-    <div class="auth-widget" v-if="user">
+    <div class="auth-widget"> 
+        <div class="auth-widget" v-if="user">
```
chukarave, line 3: There's probably a nicer way to do this that also doesn't break the styles but at this point in time i'm not aware of it

guergana: yeah it look weird.

guergana (changes requested):
It seems that you can remove the `class="auth-widget"` in the outer div and it would have the same effect. Though I am not sure I tested all possible scenarios. Could you check?

resources/js/Components/AuthWidget.vue
guergana, line 3: indentation after this line should be one to the right.

resources/js/Components/AuthWidget.vue
```diff
@@ -1,11 +1,13 @@
 <template>
-    <div class="auth-widget" v-if="user">
+    <div class="auth-widget">
```
guergana, line 2:
```suggestion
    <div class="auth-widget_wrapper"> 
```

resources/js/Components/AuthWidget.vue
```diff
@@ -39,6 +42,9 @@ export default defineComponent({
 
     a {
         white-space: nowrap;
+        &:visited {
+            color: $color-progressive;
+        }
     }
```
guergana, line 48:
```suggestion

&_wrapper {
        a {
            white-space: nowrap;
            &:visited {
                color: $color-progressive;
            }
        }
    }
```

guergana (changes requested):
Something along these lines has the same effect and doesn't nest the same class name. Lemme know if all the scenarios are covered by my suggestion :pray:

resources/js/Components/AuthWidget.vue
chukarave: This seems to do the trick, thank you!

resources/js/Components/AuthWidget.vue
```diff
@@ -1,11 +1,13 @@
...
-        </a>
-        <a  href="/auth/logout">{{ $i18n('log-out') }}</a>
+    <div class="auth-widget_wrapper"> 
+        <div class="auth-widget" v-if="user">
+            <a :href="`https://www.wikidata.org/wiki/User:${user.name}`">
+                <img src="images/user.svg" class="icon-user" /><span class="username">{{ user.name }}</span>
+            </a>
+            <a  href="/auth/logout">{{ $i18n('log-out') }}</a>
+        </div>
+        <a class="auth-widget" v-else href="/auth/login">{{ $i18n('log-in') }}</a>
```
guergana, line 9: Ah sorry, forgot to mention the class here needs to be removed, that's why we moved the css to the wrapper.

guergana (changes requested):

guergana:
Looks great. Thank you :100:

guergana (approved):
:heavy_check_mark:

PR #796, opened 2023-12-21
Remove Playground Page

guergana:
This page is no longer necessary as it was created at the beginning stage of the project to test the components.


guergana:
> If you are removing the page, you will also have to remove the `dev/playground` route from Laravel. The route is in `routes/dev.php`. In fact, I think we can remove that whole file too.

OMG! yes, i had the feeling that something was missing from my PR. this is it! thanks

itamargiv (approved):
Great job, thanks!

PR #799, opened 2023-12-21
Updates December 2023

itamargiv:
This change includes all updates that were able to be smoothly applied.


chukarave (approved):
Looks good to me. Thanks!

PR #805, opened 2023-12-28
Rewrite Results Page in Vue 3 syntax

guergana:
Bug: T354059

### DIFF
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -146,7 +146,8 @@ import LoadingOverlay from '../Components/LoadingOverlay.vue';
 import MismatchesTable from '../Components/MismatchesTable.vue';
 import Mismatch, {ReviewDecision, LabelledMismatch} from '../types/Mismatch';
 import User from '../types/User';
-import { Ref, computed, onMounted, ref } from 'vue';
+import type { Ref } from 'vue';
+import { computed, onMounted, ref } from 'vue';
 import axios from 'axios';
 
 // Run it with compat mode
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -272,15 +272,7 @@ async function send( item: string ): Promise<void> {
         return;
     }
 
-    // Casting to `any` since TS cannot understand $refs as
-    // component instances and complains about the usage of `show`
-    // See: https://github.com/vuejs/vue-class-component/issues/94
-    // Defaulting to any, as the alternative presents us with
-    // convoluted and unnecessary syntax.
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const overlay = overlayRef.value;
-
     overlay.show();
 
     // use axios in order to preserve saved mismatches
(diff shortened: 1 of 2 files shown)


resources/js/Pages/Results.vue
```diff
@@ -76,7 +76,7 @@
                         {{labels[item]}} ({{item}})
                     </a>
                 </h2>
-                <form @submit.prevent="send(item)">
+                <form @submit.prevent="send(String(item))">
```
guergana, line 79: Maybe we should declare the type of the `item` at declaration time in the `v-for` instead of casting it evertime. I am not sure if this is possible, though.

resources/js/Pages/Results.vue
```diff
@@ -136,251 +136,205 @@
...
+import { useStore } from '../store';
+import isEmpty from 'lodash/isEmpty';
+import { Head as InertiaHead } from '@inertiajs/inertia-vue3';
+import { CdxButton, CdxIcon, CdxDialog, CdxMessage, CdxCheckbox } from "@wikimedia/codex";
+import { cdxIconInfo, cdxIconArrowPrevious } from '@wikimedia/codex-icons';
+import LoadingOverlay from '../Components/LoadingOverlay.vue';
+import MismatchesTable from '../Components/MismatchesTable.vue';
+import Mismatch, {ReviewDecision, LabelledMismatch} from '../types/Mismatch';
+import User from '../types/User';
+import { Ref, computed, onMounted, ref } from 'vue';
```
chukarave, line 149: Are we sure we need both Ref and ref?

guergana: yes `Ref` is a type, `ref` is a function :)

tests/Vue/Pages/Results.spec.js
```diff
@@ -48,7 +50,7 @@ describe('Results.vue', () => {
     }
 
     beforeEach(async () => {
-        axios.put = jest.fn();
+        axios.put = jest.fn()
```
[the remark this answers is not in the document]
guergana: ah, yes, that must have slipped through in the multiple things i tried.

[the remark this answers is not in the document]
guergana: semicolon is back!

chukarave (commented):
It works fine as far as I can tell. Locally I get more console warnings than we currently have on the development branch, those seem to be Vue 3 related. Might make sense to have a look.

guergana:
> It works fine as far as I can tell. Locally I get more console warnings than we currently have on the development branch, those seem to be Vue 3 related. Might make sense to have a look.

You're right, the warnings are coming from Codex components `CdxSelect` `CdxMenuItem` and the render function which is a problem that was there before though. I don't know if some of these warnings will go away on their own once we convert the components to Vue3 as well.

resources/js/Pages/Results.vue
chukarave: It'll avoid such misunderstandings as the one I had, so i'm for it of course :grimacing:

guergana: Oh that's true. Thanks for the reminder! Will update.

hasanakg (approved):

PR #800, opened 2023-12-22
Replace tokens in pages

chukarave:
This PR is based on PR https://github.com/wmde/wikidata-mismatch-finder/pull/794 to prevent merge conflicts and as such should be merged after it.

Not many tokens to replace in pages, mostly in `Home.vue` and `Layout.vue`

Bug: T352314

PR #823, opened 2024-01-03
Replace tokens in remaining components

chukarave:
This replaces the last of the Wikit tokens found in components. 
To be merged only after #800 

Bug: T352311


resources/sass/app.scss
```diff
@@ -14,10 +14,16 @@ html {
 
 body {
     @include body-M;
-    @include body-S;
     height: 100%;
 }
 
+// TODO: This is a temporary fix until issue https://phabricator.wikimedia.org/T326900
+// is marked as resolved.
+.cdx-button .cdx-icon, .cdx-button {
```
chukarave, line 22: this is the way we found to make the mixin apply for buttons with and without Icons. This also adds the bold font weight, we might consider just adding it to the mixin itself?

sai-san: I would encourage us to add the weight as an argument in the mixin, yes.

sai-san:
UI review done. All replacements seem correct 👍

guergana (changes requested):
This one is not compiling locally for me. Same error as in the failed browser test:

```
ERROR in ./resources/sass/app.scss
Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js):
ModuleBuildError: Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
Undefined mixin.
   ╷
23 │     @include ui-text-M;
```

it seems that this PR needs to be rebased including the first PR in the chain where this token was updated

resources/sass/app.scss
guergana: The `.cdx-button .cdx-icon` part is causing the icon to be rendered in a different color, any properties assigned to the parent `.cdx-button` will be overwritten here. I suggest getting rid of the `.cdx-button .cdx-icon` part here.

guergana (changes requested):
spaces, nitpicks that will give linting problems later.

resources/sass/_typography.scss
```diff
@@ -25,14 +24,15 @@
...
+   color: $color-base;
 }
 
 @mixin ui-text-M ($font-weight: $font-weight-normal){
     font-family: $font-family-system-sans;
     font-size: $font-size-medium;
     font-weight: $font-weight;
     line-height: $line-height-xx-small;
-    color: $color-base;
+   color: $color-base;
```
guergana, line 35: this is misaligned :grimacing:

guergana (changes requested):

resources/sass/_typography.scss
chukarave: In more than one place even

guergana (changes requested):
Hi, i can't write on that line but there's also an extra double new line in lines 412-413 https://github.com/wmde/wikidata-mismatch-finder/pull/823/files#diff-b339140a5abe2cafdc4e45227556c9b8799da604d97644e62f472d5f80506725R412

guergana:
I guess more of a question for UX @sai-san : if the log in link in the AuthWidget is not supposed to have a visited state (aka purple color after visited)... is it supposed to have a visited state when it is part of the message component? 
image

chukarave:
@guergana this is a good catch :+1: the link styles are part of [T347161](https://phabricator.wikimedia.org/T347161), please add this comment there so it will not get lost

PR #826, opened 2024-01-04
Rewrite MismatchRow in Vue3 script setup syntax

guergana:
The commits from MismatchesTable #828 are included here, otherwise the tests don't pass. 

#828 is to be merged before this one. Otherwise both components will be merged under one commit if this one is merged first.

Bug: T354346


chukarave (approved):
Works well for me, thank you

PR #830, opened 2024-01-10
Improve links

chukarave:
Align user and log out buttons relatively to the language selector button.

Bug: T347161


guergana (changes requested):
It seems to me that what the current PR is doing is what was the intended use of the `align-items: center;` in this line, https://github.com/wmde/wikidata-mismatch-finder/blob/b56cac2da121ae1a18a374f176e3ffecf6a3f01c/resources/js/Components/AuthWidget.vue#L31 

but right now it doesn't seem to do anything. I tested and it seems safe to remove it from `AuthWidget` and leave it here (this PR).

guergana (approved):
Looks great! thanks!

PR #831, opened 2024-01-10
Fix textarea design errors

guergana:
Bug: T347190

### DIFF
--- a/resources/js/Components/ItemIdSearchTextarea.vue
+++ b/resources/js/Components/ItemIdSearchTextarea.vue
@@ -13,6 +13,9 @@
           :status="validationError ? validationError.type : 'default'"
           v-model="textareaInputValue"
       />
+      <template #label>
+			{{$i18n('item-form-id-input-label')}}
+		</template>
   </cdx-field>
 </template>
 
@@ -92,6 +95,7 @@ defineExpose({validate, serializeInput, validationError});
 </script>
 
 <style lang="scss">
+@import '~@wmde/wikit-tokens/dist/_variables.scss';
 
 .cdx-field__control {
     position: relative;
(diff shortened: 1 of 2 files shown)


resources/js/Pages/Home.vue
```diff
@@ -197,21 +197,5 @@ function showRandom(): void{
     .form-buttons {
         text-align: end;
     }
-
-    .cdx-field__control {
```
guergana, line 201: Removing this from `Home.vue` because it was duplicated. This code belongs in `[ItemIdSearchTextarea.vue](https://github.com/wmde/wikidata-mismatch-finder/pull/831/files#diff-ee8902c55d2443db0d269e1a35909457d37b48d775e00e61fbc631dec2b8a759)` and has already been added there.

resources/js/Components/ItemIdSearchTextarea.vue
```diff
@@ -13,6 +13,9 @@
           :status="validationError ? validationError.type : 'default'"
           v-model="textareaInputValue"
       />
+      <template #label>
+			{{$i18n('item-form-id-input-label')}}
```
chukarave, line 17: Not a merge blocker, but the label automatically gets a font-weight of 700 (on Firefox for me) which appears bolder than the Figma illustration shows. I still think this can go ahead since we will be revamping the styles we missed once tokens are on development anyway

resources/js/Components/ItemIdSearchTextarea.vue
```diff
@@ -105,6 +109,10 @@ defineExpose({validate, serializeInput, validationError});
         .cdx-progress-bar {
             width: 50%;
             margin: auto;
+
+            @media (max-width: $width-breakpoint-tablet) {
+                width: 80%;
```
chukarave, line 114: works nicely, i got a cute little progress bar on mobile. 
In this case too we will have to redo this style once tokens are in place

chukarave (approved):
tiny stuff that doesn't block merging

PR #832, opened 2024-01-10
Reassign css to proper div to remove white space below footer

guergana:
This PR solves the issue of when the TextArea's height is reduced to the minimum, white space appears under the Mismatch app footer.

Note: currently the linting is failing due to the linting rules being set for Vue2. 

Two options here:
1. Wait until T351644 changing the ruleset is updated to Vue3 and this error will be gone. The error states that only one element can be inside a `<template>` tag, but this has changed in Vue3, and multiple elements can now be added inside a template. See: https://v3-migration.vuejs.org/new/fragments.html

2. Otherwise we can disable the rule that creates the error and remove it when we update the linting rules to Vue3.

I am choosing option 2. The rules will be updated later on in the `.eslintrc.js` file.

Bug: T347190


chukarave (approved):
works well, thanks!

PR #836, opened 2024-01-15
Update linting rules to Vue3

guergana:
This PR adds `vue/vue3-strongly-recommended` to check for vue linting violations in vue3

Bug: T351644

### DIFF
--- a/app/Http/Controllers/ResultsController.php
+++ b/app/Http/Controllers/ResultsController.php
@@ -65,10 +65,10 @@ class ResultsController extends Controller
         $props = array_merge(
             [
                 'user' => $user,
-                'item_ids' => $requestedItemIds,
+                'itemIds' => $requestedItemIds,
                 // Use wikidata to fetch labels for found entity ids
                 'labels' => $wikidata->getLabels($entityIds, $lang),
-                'formatted_values' => $formattedTimeValues,
+                'formattedValues' => $formattedTimeValues,
             ],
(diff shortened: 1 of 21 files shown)


chukarave (commented):
So nice to see so many lint fixes at once :smile:

app/Http/Controllers/ResultsController.php
```diff
@@ -65,10 +65,10 @@ public function index(MismatchGetRequest $request, WikibaseAPIClient $wikidata):
         $props = array_merge(
             [
                 'user' => $user,
-                'item_ids' => $requestedItemIds,
+                'itemIds' => $requestedItemIds,
```
chukarave, line 68: Did eslint complain about a php file or was it a manual fix for consistency's sake?

chukarave: but did it lint this PHP file? I thought eslint only lints JS

guergana: these are passed as Props to the Results page here: https://github.com/wmde/wikidata-mismatch-finder/pull/836/files#diff-f0e07e4099d24711f323a8ef5a958c7723c8bdaca46379665835500790de1c3fR245

If it doesnt have the same name some tests fail.

chukarave: Ah, I see. that sounds fine

chukarave (approved):
Looks good to me :+1:

PR #838, opened 2024-01-22
Fix :visited:hover color for auth widget links

chukarave:
The Codex link mixin default of the :visited:hover style is `$color-progressive`, which is redundant for the auth widget links, since we removed :visited styles for those. 
This PR fixes the hover color specifically for the username and log in/log out buttons. 

Bug: T347161


guergana (approved):
LGTM

PR #839, opened 2024-01-22
Fix focus border outline for the store download button

chukarave:
Don't outline button on focus

Bug: T352304


guergana (approved):
Pretty straightforward, not much to review over here. LGTM!

PR #840, opened 2024-01-23
Fix broken noscript styles

chukarave:
Newly added stylelint check found some css variables have four dashes instead of two, which broke the noscript page style. 

Bug: T351644 
(this is not directly connected to this task but was found while working on it)

### DIFF
--- a/resources/sass/noscript.scss
+++ b/resources/sass/noscript.scss
@@ -9,20 +9,20 @@ main {
     font-family: $font-family-system-sans;
     font-weight: $font-weight-normal;
     color: $color-base;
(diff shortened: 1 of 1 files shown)


guergana:
@chukarave that was a good catch, since it was also breaking the stylelint code i have corrected this here as part of the stylelint ticket: https://github.com/wmde/wikidata-mismatch-finder/pull/841/files#diff-64c54febd2af0f8581a5efafb71a94a486877221624539e2752148d3c3911c3eL12 . Do you think it is worth merging it here as a separate PR? or do we just close this one and make it part of the general fixing in stylelint PR?

chukarave:
Yeah I noticed that yesterday too while reviewing the other PR. Let's close this one and leave it to be fixed with the stylelint rules :+1:

PR #841, opened 2024-01-23
Add stylelint to project

guergana:
Bug: T351644


.stylelintrc.json
```diff
@@ -0,0 +1,9 @@
+{
+    "rules": {
+        "selector-class-pattern": [null]
```
guergana, line 3: Reasoning behind the decision of disabling the `selector-class-pattern` rule:
We are loosely following the BEM naming convention and this rule expects classes to be named using kebab case. Installing a plugin to reinforce the BEM convention would mean we also need to update all the class names that are not using it and that is too much overhead for the current task.

.github/workflows/lint.yml
```diff
@@ -37,3 +37,19 @@ jobs:
 
         - name: Run i18 Validation
           run: npm run i18n:validate
+
+  style-lint:
```
chukarave, line 41: I think we might be able to avoid having two linting jobs by changing the ESlint action to run one centralized `lint` script which would run both ESlint and Stylelint (which will be configured in `package.json`). That way we won't need a separate Stylelint job here

package.json
```diff
@@ -4,6 +4,8 @@
     "scripts": {
         "lint": "eslint --ext .js,.vue resources/js",
         "lint:fix": "eslint --fix --ext .js,.vue resources/js",
+        "stylelint": "stylelint '**/*.scss' '**/*.vue'",
```
chukarave, line 7: As mentioned in the workflow file: we can add the `lint:` prefix to to anything ESlint/Stylelint related and then have the `lint` script run it all. The same goes for fix.

resources/js/Components/ItemIdSearchTextarea.vue
```diff
@@ -98,7 +98,8 @@ defineExpose({validate, serializeInput, validationError});
 </script>
 
 <style lang="scss">
-@import '~@wmde/wikit-tokens/dist/_variables.scss';
+/* stylelint-disable-next-line scss/load-no-partial-leading-underscore */
+@import '~@wmde/wikit-tokens/dist/_variables';
```
chukarave, line 102: oops, i think we can remove this. but out of the scope of this task

chukarave (commented):
Again, so nice to see so many things fixed at once :smile:

.github/workflows/lint.yml
guergana: for this we need to install a new package: https://www.npmjs.com/package/npm-run-all

chukarave: We could, if we wanted to. Otherwise i'm pretty sure we could add `&& npm run lint:stylelint` to the lint script to achieve the same purpose.

guergana: It's installed :)

package.json
guergana: Done!

resources/js/Components/ItemIdSearchTextarea.vue
guergana: Yeah, I don't think this is part of this task. Is this a leftover from the tokens?

chukarave: yeah i guess so. i'll do it as part of the grid system

chukarave (approved):
très chic

chukarave (dismissed):

PR #850, opened 2024-01-29
Port the table component from Wikit back to the Mismatch Finder

chukarave:
With the deprecation of Wikit (and until such a component is available through Codex) we are moving the table component back into the mismatch finder repo. 
This PR also includes minor component syntax updates to Vue 3.

NOTE: styles that have it, were replaced with their codex equivalent. otherwise the wikit tokens were replace with discrete values.

Bug: T348271

### DIFF
--- a/resources/js/Components/MismatchesTable.vue
+++ b/resources/js/Components/MismatchesTable.vue
@@ -1,5 +1,5 @@
 <template>
-  <wikit-table>
+  <Table>
     <thead>
       <tr>
         <th class="column-mismatch">
@@ -34,12 +34,11 @@
         :id="`mismatch-${mismatch.id}`"
       />
     </tbody>
-  </wikit-table>
+  </Table>
 </template>
 
 <script setup lang="ts">
-import { Table as WikitTable } from '@wmde/wikit-vue-components';
-
+import Table from './Table.vue';
 import MismatchRow from './MismatchRow.vue';
 
 import type { LabelledMismatch } from '../types/Mismatch';
--- a/resources/js/Components/MismatchesTable.vue
+++ b/resources/js/Components/MismatchesTable.vue
@@ -1,5 +1,5 @@
 <template>
-  <TableComponent>
+  <table-component>
     <thead>
       <tr>
         <th class="column-mismatch">
@@ -34,7 +34,7 @@
         :id="`mismatch-${mismatch.id}`"
       />
     </tbody>
-  </TableComponent>
+  </table-component>
 </template>
 
 <script setup lang="ts">
--- a/resources/js/Components/MismatchesTable.vue
+++ b/resources/js/Components/MismatchesTable.vue
@@ -1,5 +1,5 @@
 <template>
-  <Table>
+  <TableComponent>
     <thead>
       <tr>
         <th class="column-mismatch">
@@ -34,11 +34,11 @@
         :id="`mismatch-${mismatch.id}`"
       />
     </tbody>
-  </Table>
+  </TableComponent>
 </template>
 
 <script setup lang="ts">
-import Table from './Table.vue';
+import TableComponent from './TableComponent.vue';
 import MismatchRow from './MismatchRow.vue';
 
 import type { LabelledMismatch } from '../types/Mismatch';
--- a/resources/js/Components/Table.vue
+++ b/resources/js/Components/Table.vue
@@ -1,0 +1,229 @@
+<template>
+  <table
+    ref="Table"
+    :class="[
+      'wikit',
+      'wikit-Table',
+      `wikit-Table--linear-${breakpoint}`
+    ]"
+  >
+    <slot />
+  </table>
+</template>
+
+<script setup lang="ts">
+import { PropType, computed, ref } from 'vue';
+import { Breakpoint, validateBreakpoint } from '../types/Breakpoint';
+
+/**
+ * Tables display categorical information organized across rows and columns in
+ * order to facilitate the comparative analysis of data.
+ *
+ * The WiKit table component provides a wrapper around the common HTML table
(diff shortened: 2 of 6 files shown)


resources/js/Components/Table.vue
```diff
@@ -0,0 +1,229 @@
...
+ * **Example:**
+ *
+ * ```html
+ * <td data-header="Column Header">Content Here</td>
+ * ```
+ */
+	const Table = ref('Table');
+	const props = defineProps({
+		linearize: {
+			type: String as PropType<Breakpoint>,
```
chukarave, line 39: I know there's a snazzy way to do this that involves not importing `PropType` to begin with. Please share it with me!

hasanakg (commented):
I'm not sure about the component name `Table`, because it's a bit ambiguous. Nothing else from my side.

chukarave:
I know what you mean. We can't call it MismatchesTable since that's the containing component, we can't use WikitTable anymore and it's not a CodexTable... any suggestions?

guergana:
> I know what you mean. We can't call it MismatchesTable since that's the containing component, we can't use WikitTable anymore and it's not a CodexTable... any suggestions?

How about ResultsTable? or TableComponent?

resources/js/Components/MismatchesTable.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <wikit-table>
+  <Table>
```
guergana, line 2: Yeah we should def change the name because there is an html <table> tag as well

resources/js/Components/Table.vue
```diff
@@ -0,0 +1,229 @@
...
+		/**
+		* Colors
+		*/
+		background-color: #fff;
+		color: #202122;
+
+		/**
+		* Typography
+		*/
+		font-family: -apple-system,BlinkMacSystemFont,"Segoe UI",Roboto,Lato,Helvetica,Arial,sans-serif;
```
guergana, line 146: Isn't this supossed to come from some of the variables or tokens that we have recently defined?

resources/js/Components/Table.vue
```diff
@@ -0,0 +1,229 @@
...
+		font-size: 1em;
```
guergana, line 147: can this be replaced by a token?

resources/js/Components/Table.vue
```diff
@@ -0,0 +1,229 @@
...
+
+<script setup lang="ts">
+import { PropType, computed, ref } from 'vue';
+import { Breakpoint, validateBreakpoint } from '../types/Breakpoint';
+
+/**
+ * Tables display categorical information organized across rows and columns in
+ * order to facilitate the comparative analysis of data.
+ *
+ * The WiKit table component provides a wrapper around the common HTML table
```
guergana, line 22: Let's not forget to remove any mentions to Wikit ... :grimacing:

guergana (changes requested):

resources/js/Components/Table.vue
chukarave: ha! nice catch

resources/js/Components/Table.vue
chukarave: yeah i started with discrete values and then switched to tokens where possible halfway through.. I'll update them

chukarave:
> How about ResultsTable? or TableComponent?

Both are good, I think I like TableComponent better for this context

resources/js/Components/TableComponent.vue
```diff
@@ -0,0 +1,229 @@
+<template>
+  <table
+    ref="TableComponent"
+    :class="[
+      'wikit',
+      'wikit-Table',
+      `wikit-Table--linear-${breakpoint}`
```
chukarave, line 7: @guergana @hasanakg renaming those lines breaks the whole table, but I'm not actually sure why, we don't assign any explicit styles to these selectors. any ideas?

hasanakg: can you explain it a bit more @chukarave ?

resources/js/Components/TableComponent.vue
```diff
@@ -0,0 +1,229 @@
...
+			font-weight: $font-weight-bold;
+		}
+
+		// Hide empty cells
+		td:empty {
+			display: none;
+		}
+	}
+
+	.TableComponent {
```
hasanakg, line 119: This selector can be renamed to `table-component`

resources/js/Components/TableComponent.vue
hasanakg: You can use ```['table-component', `table-component--linear${breakpoint}`]```, right now this doesn't use the style from `TableComponent.vue`

hasanakg (changes requested):

resources/js/Components/TableComponent.vue
chukarave: nice! I think you got it

resources/js/Components/MismatchesTable.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <wikit-table>
+  <TableComponent>
```
hasanakg, line 2: I think we use `kebab-case` for component tags, instead of `UpperCamelCase`, we can continue to do so for the sake of consistency. What do you think?

guergana: I agree. I didn't know it was optional. :confused:

chukarave: Done :+1:  We could consider a linting rule for this if we don't want it to be optional.

hasanakg: We should have a day-long meeting to decide our linting rules I think :D

hasanakg (approved):

PR #849, opened 2024-01-24
Adapt layout to codex's grid system

chukarave:
This PR adapts the layout to the new Codex values, Mostly breakpoints and max-width values. 

Bug: T348729

PR #861, opened 2024-01-30
Add max-width for store unintentionally removed

chukarave:
This style was removed as part of the grid changes (#849), but it is required for the store page. Added a hint to avoid further misunderstanding


guergana (approved):
LGTM!

PR #864, opened 2024-01-30
Add overwritten button text span items

chukarave:
This is a redo of PR #837 that got partially removed by PR #841 unintentionally.  
It extracts the text out of all icon buttons into a nesting <span> element. It also adds a missing brace in the styles.

Bug: [T352304](https://phabricator.wikimedia.org/T352304)


guergana (approved):
Thanks!

PR #865, opened 2024-01-30
Remove Wikit MenuItem import

guergana:
Bug: T356172

PR #873, opened 2024-02-02
Override list item styling for select menu items

chukarave:
The menu items on the results view ended up showing a margin inherited from `app.scss` 's style definition for `<li>` elements (meant for lists like the FAQ and instructions).
This PR overrides the inherited style definitions.

Bug: T348271


chukarave:
This is already handled as part of #865! closing this

PR #874, opened 2024-02-02
Add last layout and grid improvements

chukarave:
- Media queries added to footer  
- Buttons should not show ellipsis as long as they have room to show button text

PR: T348729

### DIFF
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -15,6 +15,7 @@
           {{ $i18n('results-page-title') }}
         </h2>
         <cdx-button
+          class="heading-button"
           id="instructions-button"
           weight="quiet"
           action="progressive"
@@ -396,10 +397,6 @@ function _handleConfirmation(){
 
     // to avoid visual grouping with .description-section
     margin-bottom: var(--dimension-layout-xsmall);
-}
-
-#instructions-button {
-    flex-shrink: 0;
 }
 
 h2 {
--- a/resources/js/Pages/Results.vue
+++ b/resources/js/Pages/Results.vue
@@ -398,6 +398,10 @@ function _handleConfirmation(){
     margin-bottom: var(--dimension-layout-xsmall);
 }
 
+#instructions-button {
+    flex-shrink: 0;
+}
+
 h2 {
     a {
         font-weight: bold;
--- a/resources/sass/app.scss
+++ b/resources/sass/app.scss
@@ -381,6 +381,10 @@ dl.import-meta .download-csv {
     }
 }
 
+.heading-button, .random-mismatches {
+    flex-shrink: 0; // don't show ellipsis as long as the text fits
+}
+
 .svg {
     &-white-fill {
         filter: invert(1);
--- a/resources/sass/app.scss
+++ b/resources/sass/app.scss
@@ -381,7 +381,7 @@ dl.import-meta .download-csv {
     }
 }
 
-.heading-button, .random-mismatches {
+.heading-button {
     flex-shrink: 0; // don't show ellipsis as long as the text fits
 }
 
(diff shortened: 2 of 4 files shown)


resources/js/Pages/Results.vue
```diff
@@ -398,6 +398,10 @@ function _handleConfirmation(){
     margin-bottom: var(--dimension-layout-xsmall);
 }
 
+#instructions-button {
+    flex-shrink: 0;
```
guergana, line 402: is there a reason why this is not being set for the whole app? instead of individually for Home and Results page? or in other words... which instances of `cdx-button` need the ellipsis?

guergana: I see the design now. So only one button should have an ellipsis? :thinking:

guergana (commented):
just a question about the ellipsis in the buttons, the rest looks great.

resources/js/Pages/Results.vue
chukarave: I was debating this exact point myself. I decided to play it safe and go with every button that has a heading next to it, where we originally added the gap between button and heading text ("More information" button, "Random Mismatches" and the instructions button). 

I'm not against adding this globally, i just wasn't sure about possible side effects

guergana: This seems a perfect occasion to ask @sai-san

chukarave:
@hasanakg can you specify what it is about the `flex-shrink`?

hasanakg:
@chukarave the one @guergana commented, I think we should wait @sai-san for that one?

chukarave:
@hasanakg The question was - where would this code be easier to maintain/contain less side effects, so not strictly a UX question. @guergana suggested adding a selector to said buttons, which I think is a good compromise, so I did.

resources/sass/app.scss
```diff
@@ -381,6 +381,10 @@ dl.import-meta .download-csv {
     }
 }
 
+.heading-button, .random-mismatches {
+    flex-shrink: 0; // don't show ellipsis as long as the text fits
+}
+
```
chukarave, line 387: This didn't work as long as it was inside the `.cdx-button` selector, i'm not sure why. I extracted it and it works like this.

guergana:
> @chukarave the one @guergana commented, I think we should wait @sai-san for that one?

@hasanakg `flex-shrink` was suggested by Sarai, here: https://phabricator.wikimedia.org/T348729#9503722 and it works :)

resources/sass/app.scss
```diff
@@ -381,6 +381,10 @@ dl.import-meta .download-csv {
     }
 }
 
+.heading-button, .random-mismatches {
```
guergana, line 384: you should add the `heading-button`class  to `.random-mismatches` in `Home.vue` and have a single selector here. or is random-mismatches not a heading button? 

```
<cdx-button
          class="random-mismatches heading-button"
          weight="normal"
          @click="showRandom()"
          :disabled="loading"
        >
```

guergana, line 384:
```suggestion
.heading-button {
```

guergana (changes requested):

resources/sass/app.scss
guergana: ahhhh, this is what you mean.... i added a comment suggesting to do this.. will check again and get back to you

resources/sass/app.scss
guergana: This works for me: 
image

chukarave: no, I think you're right. I just wasn't sure how to add it

hasanakg (approved):

guergana (approved):
LGTM! Thanks

PR #875, opened 2024-02-05
Remove compat build

hasanakg:
Bug: T351812

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -22,8 +22,6 @@ module.exports = {
     // "import Counter from '@/Counter.vue'"
     // (no need for a full path)
     "moduleNameMapper": {
-        "^vue$": "@vue/compat",
-        '^@vue/composition-api$': '@vue/compat',
         '^@wmde/wikit-vue-components$':
             '@wmde/wikit-vue-components/dist/wikit-vue-components-vue3compat.common.js',
         '^wikit-dist(.*)$': "<rootDir>/node_modules/@wmde/wikit-vue-components/dist$1",
@@ -33,18 +31,5 @@ module.exports = {
     // Further info: https://test-utils.vuejs.org/migration/#-vue-vue3-jest-jest-28
     "testEnvironmentOptions": {
         "customExportConditions": ["node", "node-addons"],
-    },
-    // For Vue migration build
-    // Add compat config to test as well
-    "globals": {
-        "vue-jest": {
-            "compilerOptions": {
-                compatConfig: {
-                    MODE: 3,
-                    COMPILER_V_ON_NATIVE: true,
-                    COMPILER_V_BIND_SYNC: true
-                }
-            }
-        }
     }
 }
--- a/package.json
+++ b/package.json
@@ -59,7 +59,6 @@
     "dependencies": {
         "@inertiajs/inertia": "^0.11.0",
         "@inertiajs/inertia-vue3": "^0.6.0",
-        "@vue/compat": "^3.3.8",
         "@wmde/wikit-tokens": "^2.1.0-alpha.15",
         "@wmde/wikit-vue-components": "^2.1.0-alpha.16",
         "date-fns": "^3.3.1",
--- a/resources/js/Components/LoadingOverlay.vue
+++ b/resources/js/Components/LoadingOverlay.vue
@@ -14,8 +14,8 @@ import { ref } from 'vue';
 import { CdxProgressBar } from "@wikimedia/codex";
 
 const props = withDefaults(defineProps<{
-	delay: number
-	visible: boolean
+	delay?: number
+	visible?: boolean
 }>(), {
 	delay: 250,
 	visible: false
--- a/tests/Vue/Components/MismatchRow.spec.js
+++ b/tests/Vue/Components/MismatchRow.spec.js
@@ -275,7 +275,7 @@ describe('MismatchesRow.vue', () => {
 
         await wrapper.find('.full-description-button').trigger('click');
 
-        const dialog = wrapper.find('.full-description-dialog .cdx-dialog');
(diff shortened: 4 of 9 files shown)


package.json
```diff
@@ -59,7 +59,6 @@
     "dependencies": {
         "@inertiajs/inertia": "^0.11.0",
         "@inertiajs/inertia-vue3": "^0.6.0",
-        "@vue/compat": "^3.3.8",
         "@wmde/wikit-tokens": "^2.1.0-alpha.15",
         "@wmde/wikit-vue-components": "^2.1.0-alpha.16",
```
chukarave, line 63: I think we can remove lines 62 and 63 too

resources/js/Components/LoadingOverlay.vue
```diff
@@ -14,8 +14,8 @@ import { ref } from 'vue';
 import { CdxProgressBar } from "@wikimedia/codex";
 
 const props = withDefaults(defineProps<{
-	delay: number
-	visible: boolean
+	delay?: number
+	visible?: boolean
```
chukarave, line 18: Isn't making these optional kind of risky? If we somehow don't pass `visible` the `show()` method breaks. Is there a way around this that leaves them required?

hasanakg: They have default values, therefore they won't be `null` or `undefined`

package.json
hasanakg: I think removing `wikit` is not in the scope of this ticket, that's why I left these lines.

jest.config.js
```diff
@@ -22,8 +22,6 @@ module.exports = {
     // "import Counter from '@/Counter.vue'"
     // (no need for a full path)
     "moduleNameMapper": {
-        "^vue$": "@vue/compat",
-        '^@vue/composition-api$': '@vue/compat',
         '^@wmde/wikit-vue-components$':
             '@wmde/wikit-vue-components/dist/wikit-vue-components-vue3compat.common.js',
         '^wikit-dist(.*)$': "<rootDir>/node_modules/@wmde/wikit-vue-components/dist$1",
```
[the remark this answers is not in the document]
guergana: > I think removing `wikit` is not in the scope of this ticket, that's why I left these lines.

@hasanakg You are right that this is not the scope of this ticket but i think there is no ticket to remove wikit yet, we should do a cleanup afterwards.

resources/js/Components/LoadingOverlay.vue
guergana: > They have default values, therefore they won't be `null` or `undefined`

I also had this issue, i was getting a TS warning or error because of this at some point but i dont see it anymore and they do have defaults, but my question is... is this part of the scope of this ticket? :grimacing:

hasanakg: Valid question :D Vue documentation says that you should remove all the warnings before removing the compat build, I took it as a sign :D

guergana: > Valid question :D Vue documentation says that you should remove all the warnings before removing the compat build, I took it as a sign :D

ah ok, so you got the warning.. i also saw it. yes, we shouldn't merge code with warnings :)

hasanakg:
> The `package-lock.json` file is missing from the changes.

Fixed now.

guergana (changes requested):
As part of this task we should remove any references to the compat build and extra imports / hacks we used, like these two:

```
// Run it with compat mode
// https://v3-migration.vuejs.org/breaking-changes/v-model.html
CdxTextArea.compatConfig = {
    ...CdxTextArea.compatConfig,
    COMPONENT_V_MODEL: false,
};
```
in https://github.com/wmde/wikidata-mismatch-finder/blob/development/resources/js/Components/ItemIdSearchTextarea.vue#L33

and 

```
// Run it with compat mode
// https://v3-migration.vuejs.org/breaking-changes/v-model.html
CdxCheckbox.compatConfig = {
    ...CdxCheckbox.compatConfig,
    COMPONENT_V_MODEL: false,
};
```
in https://github.com/wmde/wikidata-mismatch-finder/blob/development/resources/js/Pages/Results.vue#L202

resources/js/Components/LoadingOverlay.vue
guergana: > Isn't making these optional kind of risky? If we somehow don't pass `visible` the `show()` method breaks. Is there a way around this that leaves them required?

@chukarave the only instance of the loading overlay we are using in the app is not passing the props and this is the reason we are getting the warnings `<loading-overlay ref="overlayRef" />` https://github.com/wmde/wikidata-mismatch-finder/blob/development/resources/js/Pages/Results.vue#L3 so yeah, the right call is to make them optional, since they are not being passed anyway

tests/Vue/Components/MismatchRow.spec.js
```diff
@@ -275,7 +275,7 @@ describe('MismatchesRow.vue', () => {
 
         await wrapper.find('.full-description-button').trigger('click');
 
-        const dialog = wrapper.find('.full-description-dialog .cdx-dialog');
```
guergana, line 278: everything is working flawlessly, but how is this related to the compat build??? :thinking:

hasanakg: TBH, I have no idea. The old one works with compat build, but needed to update to the new one to make it work without compat build.

guergana (approved):
Works flawlessly, thanks

resources/js/Components/LoadingOverlay.vue
chukarave: Right those were exactly the warnings we were getting. Nice to have a clean console :raised_hands: well done @hasanakg

PR #877, opened 2024-02-06
Don't show outline on logo

chukarave:
This reinstates the outline style overwritten during the token chain merge

Bug: T347161


guergana:
Oh, noo, this again?? how many things did we overwrite? :fearful:

guergana (approved):
LGTM!

PR #876, opened 2024-02-05
T356618: remove broken tool reference from layout

andrewtavis-wmde:
Hi there :wave:

As a part of exploring WMDE tools that are currently broken in [T356618](https://phabricator.wikimedia.org/T356618), I found that there's a reference to one of them in the Mismatch Finder Layout. The broken link is the following:

https://wikidata-analytics.wmcloud.org/app/CuriousFacts

This commit removes the reference. Please let me know if there's anything else I should do!

PR #878, opened 2024-02-06
Remove curious facts link from footer

chukarave:
Quoting @andrewtavis-wmde from PR #876:
"Hi there 👋
As a part of exploring WMDE tools that are currently broken in [T356618](https://phabricator.wikimedia.org/T356618), I found that there's a reference to one of them in the Mismatch Finder Layout. The broken link is the following:

https://wikidata-analytics.wmcloud.org/app/CuriousFacts

This commit removes the reference. Please let me know if there's anything else I should do!"

The PR does not manage to run CI (presumably due to some git permission issue?) so I am uploading it myself instead. 

Bug: T356618

### DIFF
--- a/resources/js/Pages/Layout.vue
+++ b/resources/js/Pages/Layout.vue
@@ -71,11 +71,6 @@
         </a>
       </p>
       <p>
-        <a href="https://wikidata-analytics.wmcloud.org/app/CuriousFacts">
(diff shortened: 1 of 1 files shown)


guergana (approved):
Very curious fact that the link is not working...

PR #880, opened 2024-02-06
Ignore already reviewed mismatches on reupload

guergana:
Bug: T329631

### DIFF
--- a/app/Jobs/ImportCSV.php
+++ b/app/Jobs/ImportCSV.php
@@ -47,9 +47,16 @@ class ImportCSV implements ShouldQueue
         $filepath = Storage::disk('local')
             ->path('mismatch-files/' . $this->meta->filename);
 
-        DB::transaction(function () use ($reader, $filepath) {
+        $db_mismatches_by_current_user = DB::select(
+            'select * from users JOIN mismatches ON mismatches.user_id = users.id
+            WHERE mw_userid = :mw_userid',
+            ['mw_userid' =>$this->meta->user->mw_userid]
+        );
 
-            $reader->lines($filepath)->each(function ($mismatchLine) {
+        DB::transaction(function () use ($reader, $filepath, $db_mismatches_by_current_user) {
+
+
+            $reader->lines($filepath)->each(function ($mismatchLine) use ($db_mismatches_by_current_user) {
 
                 // TODO: question, should we list all columns one by one or try to do something like
                 // Schema::getColumnListing('mismatches'); // where is mismatches is the table name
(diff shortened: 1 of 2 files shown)


app/Jobs/ImportCSV.php
```diff
@@ -46,16 +46,63 @@ public function handle(CSVImportReader $reader)
...
+                &$where_clauses
+            ) {
+
+                $new_mismatch = Mismatch::make($mismatchLine);
+                $new_mismatches[] = $new_mismatch;
+                $collection = collect($new_mismatch->getAttributes());
+                $collection->forget('review_status');
+                $newArray = [['review_status', '!=', 'pending']];
+                $collection->map(function ($item, $key) use (&$newArray) {
+                    if ($key != 'type') { // key can be empty in the file but in the db always has statement by default
```
chukarave, line 72: I guess we shouldn't leave out the type column completely? De facto the mismatches with the empty type column keep getting reuploaded every time. 
Another weird-but-not-impossible use case is comparing two mismatches where the only difference is the type column (one is statement one is qualifier). We would be wrongfully counting it as a duplicate.

app/Jobs/ImportCSV.php
```diff
@@ -46,16 +46,63 @@ public function handle(CSVImportReader $reader)
...
+                        $newArray[] = [$key, $item];
```
chukarave, line 73: Are we sure we need `$newArray`? instead of assigning directly to the `$where_clauses` in the loop, that is. Would it effect performance somehow?

app/Jobs/ImportCSV.php
```diff
@@ -46,16 +46,63 @@ public function handle(CSVImportReader $reader)
...
+            ) {
+
+                $new_mismatch = Mismatch::make($mismatchLine);
+                $new_mismatches[] = $new_mismatch;
+                $mismatch_attributes = collect($new_mismatch->getAttributes());
+                $mismatch_attributes->forget('review_status');
+                $where_clause = [['review_status', '!=', 'pending']];
+                $mismatch_attributes->map(function ($attribute, $key) use (&$where_clause) {
+                    // key can be empty in the file but in the db always has statement by default
+                    if ($key != 'type' || $attribute != null) {
```
chukarave, line 72: I'm not sure why, but leaving the type out makes mismatches with an empty `type` column get reuploaded as duplicates. I think we want to avoid that

guergana: Yes, that's why this was added. Because the type column will be converted to statement automatically when empty. So we don't want to take this column from the CSV into account.

chukarave: Yes, that's clear. What happens in practice, is that (at least locally for me) the one mismatch in the example file that has no type keeps getting reuploaded, although it has been reviewed. 
We should also consider the (improbable yet not impossible) case where the same mismatch is uploaded but with a qualifier type, but we leave out the type so we will consider it a duplicate, which would be wrong.

hasanakg: @chukarave this shouldn't be the case, I will try this on my local as well

chukarave:
@hasanakg @guergana it works great now! Do you think we need a test for the empty type use case?

chukarave (approved):
me gusta :+1:

PR #883, opened 2024-02-19
Allow the language selector to detect autonym by exonym

chukarave:
This PR adds the ability to search for a language name in any language and have the autonym suggested (same behavior as Wikidata). 

Bug: T353727

### DIFF
--- a/resources/js/Components/LanguageSelector.vue
+++ b/resources/js/Components/LanguageSelector.vue
@@ -74,7 +74,10 @@ const shownLanguages = computed<Language[]>(() => {
 
 function onInput(searchedLanguage: string): void {
 	searchInput.value = searchedLanguage;
-    debouncedApiLanguageSearch(searchInput.value);
+    if (searchInput.value) {
+        debouncedApiLanguageSearch(searchInput.value);
+    }
+
 	highlightedIndex.value = 0;
 }
 
--- a/resources/js/Components/LanguageSelector.vue
+++ b/resources/js/Components/LanguageSelector.vue
@@ -47,7 +47,7 @@ import languageData from "@wikimedia/language-data";
 const searchInput: Ref<string> = ref('');
 const highlightedIndex: Ref<number> = ref(-1);
 const closeUrl = ref(closeUrlSvg);
-const apiSearchInput: Ref<string> = ref('');
+const apiSearchInput = ref(['']);
 
 const input = ref<InstanceType<typeof LanguageSelectorInput> | null>(null);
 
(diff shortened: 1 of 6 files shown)


hasanakg (commented):
Everything works great! One exception: I think we need to have a debounce period for API searches, right now we send an API call immediately after key down, but we need to call the API if the user doesn't write for some period like 300ms.

resources/js/Components/LanguageSelector.vue
```diff
@@ -73,22 +74,22 @@ const shownLanguages = computed<Language[]>(() => {
...
             params: {
-                search: inputValue,
+                search: debouncedInputValue,
                 origin: '*' // avoid CORS console errors
             }
         }).then((response) => {
             apiLanguageCodes.value = Object.keys(response.data.languagesearch);
         });
-}
+}, 200);
```
hasanakg, line 92: For me, 200ms is good, but should we give a heads-up to the UX team?

resources/js/Components/LanguageSelector.vue
```diff
@@ -73,22 +74,22 @@ const shownLanguages = computed<Language[]>(() => {
 
 function onInput(searchedLanguage: string): void {
 	searchInput.value = searchedLanguage;
```
hasanakg, line 76: If I write something and then remove it, `searchInput.value` is empty and the code makes an API call with an empty language code string.

hasanakg (changes requested):
We need to solve the empty input problem that I mentioned in the code comments.

resources/js/Components/LanguageSelector.vue
chukarave: Not good! should work better now

resources/js/Components/LanguageSelector.vue
chukarave: I discussed this last week with Sarai and it was decided there is not need to plan changes to the UX flow until deemed necessary in product verification. see [here](https://phabricator.wikimedia.org/T353727#9554935).

hasanakg (approved):

guergana:
@chukarave as much as I would have liked to come up with the solution for the svg-jest mock transformer by myself, i actually found it in this thread, in case you wanna take a look: https://github.com/jpkleemans/vite-svg-loader/issues/35#issuecomment-1341893590

chukarave:
@guergana nice! thanks. it's pretty straight forward, I like that it doesn't require installing any svg tansformer packages.

hasanakg (approved):

PR #972, opened 2026-09-02
Chore: update composer dependencies

yerdua:
This commit updates the sail runtime from php 7.4 to 8.1, due to 7.4
relying on the no-longer-available postgres "focal" release. This was
a necessary prerequisite to being able to set up a local dev environment
using the quickstart instructions and to update the composer
dependencies.

A note for reviewing this PR: it is now expected that `browser-tests`
will fail. The cause is that larave/dusk versions prior to 7.0 use a
now-obsolete url to find the latest chrome version and download it.
These tests will continue to fail until laravel itself is updated to
9.0 or higher ([T436831](https://phabricator.wikimedia.org/T436831))

Updated dependencies:

- fruitcake/laravel-cors     2.2.0 -> 3.0.0
  3.0.0 is the last version of this package before it was abandoned
  because its functionality is included in laravel core from 9.2 [1].
  Since we're still on laravel 8, we can't yet remove the dependency.
  From 2.2.0 and 3.0.0, support for some older versions of php, laravel
  and phpunit were dropped.

- inertiajs/inertia-laravel  0.6.11 -> 1.3.4
  bug fixes, removed support for laravel 6 & 7

- kevinrob/guzzle-cache-middleware  5.1.0 -> 7.0
  drop support for php versions <8.1

- laravel/tinker    2.8 -> 3.0.2
  performance improvements, error handling

- taavi/laravel-socialite-mediawiki  1.4 -> 1.7.1
  in https://github.com/wmde/wikidata-mismatch-finder/pull/834, because of an outage, the source of this package was set to a
  github clone, intended to be temporary. This outage is long since
  resolved, so the github workaround is removed.
  Between 1.4 and 1.7.1, there were dependecy updates, support for newer
  versions of laravel an php, and minor code cleanup.

- squizlabs/php_codesniffer  3.7.2 -> 3.13.6
  was abandoned by squizlabs and then picked up by PHPCSStandards. There
  were bug, performance and security fixes, as well as supporting newer
  versions of php. Version 4.0 introduces breaking changes that result
  in linting warnings and errors. This upgrade will be done in [T436541](https://phabricator.wikimedia.org/T436541)

[1]: https://github.com/fruitcake/laravel-cors#this-package-is-deprecated-because-all-supported-laravel-versions-now-include-the-cors-middleware-in-the-core


composer.json
thiemowmde: You managed to update laravel/sail to "^1.25" in a0fe333c. Was this intentionally skipped here?

thiemowmde (approved):

composer.json
yerdua: good catch. That was not intentional, and I can push a new version of the commit including it later today.

thiemowmde:
I have never worked with this codebase before, unfortunately. But what I can see in this patch looks really good to me. :+1: I believe we can merged it as it is. The failing browser tests are explained in [T436831](https://phabricator.wikimedia.org/T436831). What would you suggest as a way forward here? One possible solution I can see is to force-merge this as it is (none of the updates should have a visible effect on production, as far as I can see) and resolve the remaining issue(s) in the next patch(es), as outlined in T436831. Would that work?

Should we also mention [T435631](https://phabricator.wikimedia.org/T435631) in the commit message?

yerdua:
the latest version now includes updating `laravel/sail` to `^1.25`, and the issue with browser tests is linked in the commit message.

I think the path forward is to force merge this, attempt a new deploy and see if the tool still runs. And if not, rollback.

thiemowmde (dismissed):

